PR #5, opened 2019-07-31
fixed go-lint

tnakagawa:
Fixed the source being noted with go-lint.

golint utreexo/*.go


adiabat:
ok, we can try to make go-lint happy :)

PR #7, opened 2019-08-02
Change PopCount logic

tnakagawa:
This PR changes the PopCount logic to fast logic.
It will be about 50 times faster.
It is based on wikipedia.
https://en.wikipedia.org/wiki/Hamming_weight

However, if the method is not used frequently, there is no need to change it.


adiabat:
Thanks but sorry, I think this isn't worth the trade-off in complexity.  This is quite a bit harder to read and understand for a function that is called only once per block, and only takes a few nanoseconds.

Looking at this loop though, it does seem it could be made more efficient without any more advanced algorithms.  Instead of a fixed loop through all 64 bits, a check for non-zero `i` would be faster and eliminate the `j` variable.

PR #8, opened 2019-08-20
add store and restore for Pollard

tnakagawa:
Pollard of store and restore code.
Stores the number of leaves, all the hash of leaves, and all the top hashes.
The correctness of the data is determined by restoring Pollard and matching the top hash.

If this PR is unnecessary, it will close immediately.

### DIFF
--- a/utreexo/pollarutil.go
+++ b/utreexo/pollarutil.go
@@ -100,7 +100,7 @@ func (p *Pollard) store(w io.Writer) error {
 	// TODO: Need to lock update
 	// Write the number of leaves.
 	num := make([]byte, 8)
-	binary.LittleEndian.PutUint64(num, p.numLeaves)
+	binary.BigEndian.PutUint64(num, p.numLeaves)
 	w.Write(num)
 	// Write the hash values.
 	_, hs := getTopsReverse(p.numLeaves, p.height())
@@ -144,7 +144,7 @@ func restore(r io.Reader) (*Pollard, error) {
 	if err != nil {
 		return nil, err
 	}
-	numLeaves := binary.LittleEndian.Uint64(num)
+	numLeaves := binary.BigEndian.Uint64(num)
 	// Restore a Pollard.
 	p := &Pollard{}
 	for i := uint64(0); i < numLeaves; i++ {
--- a/utreexo/pollarutil.go
+++ b/utreexo/pollarutil.go
@@ -1,5 +1,13 @@
 package utreexo
 
+import (
+	"bufio"
+	"encoding/binary"
+	"fmt"
+	"os"
+	"reflect"
+)
+
 // Pollard is the sparse representation of the utreexo forest, using
 // binary tree pointers instead of a hash map.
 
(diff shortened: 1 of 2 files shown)


adiabat:
Hi @tnakagawa -- thanks for this PR and I do think this functionality is necessary.  There are some revisions I'd like to suggest before merging.  One of them may be a substantial change.

The easy changes first:
- I'd like to stick to big endian throughout serialization in this library as this will be used as a network protocol.

- Instead of having `store` and `restore` functions which take a filepath, similar functions of `toBytes` and `fromBytes` which output a `[]byte` and take in a `[]byte` are more convenient, because then you can also send them over a network interface.  In this case the Pollard may be very large, in which case keeping it in RAM first isn't a good idea.  In that case, functions that use `io.Reader` and `io.Writer` would be more appropriate as it could be adapted to writing to disk or writing to a network interface.

Those two changes seem easy enough.  The change that seems more complex:

- `store` assumes both children of a node are populated:
```
w.Write(node.niece[0].data[:])
w.Write(node.niece[1].data[:])
```
this will crash if the node has a nil niece.  Right now this works in `TestStoreAndRestore()` because every node added is kept (using `p.addOne(r, true)` to ensure a fully populated pollard).  In general this is not the case, and the pollard will be sparse.  A quick fix for this would be to say that the `store` and `restore` functions only operate on fully populated pollards for now, and we can figure out sparse pollards later; in that case return an error when trying to store a sparse pollard. 

Alternatively, have checks for nil nieces at every step, so that it can support serialization of sparse pollards.

Thanks and let me know if I can be of any help!

tnakagawa:
Hi @adiabat .

I update two easy changes.
- change file path to io.Writer or io.Reader
- change LittleEndian to BigEndian

tnakagawa:
I'm thinking of store and restore Pollad including `p.addOne(r, false)`.
It will take time ...

tnakagawa:
Are the following results normal?
I think L and R are the opposite results for `TestAddOne2` and `TestAddOne3`.

---

```golang
func TestAddOne1(t *testing.T) {
	h1 := Hash{0x01}
	h2 := Hash{0x02}
	p := &Pollard{}
	p.addOne(h1, true)
	p.addOne(h2, true)
	top := p.tops[0]
	t.Logf("T:%x", top.data)
	t.Logf("L:%x", top.niece[0].data)
	t.Logf("R:%x", top.niece[1].data)
}
```

```
T:78e5e778eeb4f10f474ee0b96561667de1d0a14bea82cb81bf34ce2a14d9fdf2
L:0100000000000000000000000000000000000000000000000000000000000000
R:0200000000000000000000000000000000000000000000000000000000000000
```

---

```golang
func TestAddOne2(t *testing.T) {
	h1 := Hash{0x01}
	h2 := Hash{0x02}
	p := &Pollard{}
	p.addOne(h1, true)
	p.addOne(h2, false)
	top := p.tops[0]
	t.Logf("T:%x", top.data)
	t.Logf("L:%v", top.niece[0])
	t.Logf("R:%x", top.niece[1].data)
}
```

```
T:78e5e778eeb4f10f474ee0b96561667de1d0a14bea82cb81bf34ce2a14d9fdf2
L:<nil>
R:0200000000000000000000000000000000000000000000000000000000000000
```

---

```golang
func TestAddOne3(t *testing.T) {
	h1 := Hash{0x01}
	h2 := Hash{0x02}
	p := &Pollard{}
	p.addOne(h1, false)
	p.addOne(h2, true)
	top := p.tops[0]
	t.Logf("T:%x", top.data)
	t.Logf("L:%x", top.niece[0].data)
	t.Logf("R:%v", top.niece[1])
}
```

```
T:78e5e778eeb4f10f474ee0b96561667de1d0a14bea82cb81bf34ce2a14d9fdf2
L:0100000000000000000000000000000000000000000000000000000000000000
R:<nil>
```

---

```golang
func TestAddOne4(t *testing.T) {
	h1 := Hash{0x01}
	h2 := Hash{0x02}
	p := &Pollard{}
	p.addOne(h1, false)
	p.addOne(h2, false)
	top := p.tops[0]
	t.Logf("T:%x", top.data)
	t.Logf("L:%v", top.niece[0])
	t.Logf("R:%v", top.niece[1])
}
```

```
T:78e5e778eeb4f10f474ee0b96561667de1d0a14bea82cb81bf34ce2a14d9fdf2
L:<nil>
R:<nil>
```

---

adiabat:
@tnakagawa -
The behavior in `p.addOne()` may be a bug.  I think that was what I intended but then wanted to change it.  The idea is that if you want to remember 0x01, you don't have to store 0x01 in the pollard; you only have to store 0x02, because the *proof* for 0x01 is 0x02.  And the opposite; the proof for 0x02 is 0x01.  

So this makes some sense, but I thought I changed it to remember both if either is flagged as memorable.  That change may be in my undo branch, which I should try to merge in to master soon.

Even if it's not the correct behavior for minimizing sent hashes, it should still work for serializing a sparse tree.

tnakagawa:
Thank you @adiabat .

I would like to think how to save and restore without using `p.addOne()`.
Would you prefer to change the method name and add it to this PR or use another PR?

I have a question.
There is data under the leaves of a branch with only one leaf, do you need to store it too?

Examples of `TestAddOne2` and `TestAddOne3` are as follows:

---

```golang
func TestAddOne2(t *testing.T) {
	h1 := Hash{0x01}
	h2 := Hash{0x02}
	p := &Pollard{}
	p.addOne(h1, true)
	p.addOne(h2, false)
	top := p.tops[0]
	t.Logf("T:%x", top.data)
	t.Logf("L:%v", top.niece[0])
	t.Logf("R:%x", top.niece[1].data)
	t.Logf("R[0]:%x", top.niece[1].niece[0].data)
	t.Logf("R[1]:%v", top.niece[1].niece[1])
}
```

```
T:78e5e778eeb4f10f474ee0b96561667de1d0a14bea82cb81bf34ce2a14d9fdf2
L:0100000000000000000000000000000000000000000000000000000000000000
L[0]:0200000000000000000000000000000000000000000000000000000000000000
L[1]:<nil>
R:<nil>
```

---

```golang
func TestAddOne3(t *testing.T) {
	h1 := Hash{0x01}
	h2 := Hash{0x02}
	p := &Pollard{}
	p.addOne(h1, false)
	p.addOne(h2, true)
	top := p.tops[0]
	t.Logf("T:%x", top.data)
	t.Logf("L:%x", top.niece[0].data)
	t.Logf("L[0]:%x", top.niece[0].niece[0].data)
	t.Logf("L[1]:%v", top.niece[0].niece[1])
	t.Logf("R:%v", top.niece[1])
}
```

```
T:78e5e778eeb4f10f474ee0b96561667de1d0a14bea82cb81bf34ce2a14d9fdf2
L:0100000000000000000000000000000000000000000000000000000000000000
L[0]:0200000000000000000000000000000000000000000000000000000000000000
L[1]:<nil>
R:<nil>
```

adiabat:
using `p.addOne()` for testing is OK, it's just that only setting remember to true is unrealistic as in many cases pollard will have a sparse representation of the hash forest, with many empty nodes.  For example, this is a possible configuration:

```
14 
|---------------\
12              13 
|-------\       |-------\
        09              11 
|---\   |---\   |---\   |---\
00  01          04   

```

In this example, both 12 and 13 only have right pointers.  So `node12.niece[0] == nil`.  Despite this, `node12.niece[1] == node11`, and node13.niece[1] == node09`.  9 points to 00 and 01, and 11 points only to 04.

This is useful if 00 and 01 are going to be removed soon, but 05, 06 will not be removed.  We don't need to keep track of what 05 and 06 are, and since we don't need 05 and 06, 10 is also not needed.

Being able to save and load sparse forests to disk would be useful as we would reduce the proofs later downloaded.  It's good to have, but not 100% needed for the first version since we could just re-populate in ram from downloaded proofs, so a save / load function that only deals with fully populated pollards would be OK to start with.  (but if you get sparse save / load to work, great!)

tnakagawa:
Thank you @adiabat 
This PR ends.
Think about store and restore spaced Pollards separately.

PR #13, opened 2019-10-10
Combined blockparser, txottl, and ibdsim. Added flags for them

kcalvinalvin:
Combined the commands so you don't have to build them separately. Added flags with github.com/spf13/cobra.

created commands/ directory inside utreexo/cmd/

go build inside utreexo/cmd/commands/ and that binary contains blockparser, txottl, and ibdsim.


adiabat:
Thanks, will test this out a bit more tomorrow!  A couple requests:

Can we use the standard library flags function?  I'm more familiar with it and haven't used cobra.  I guess it's no big deal because it looks similar though if you want to stick with cobra.

Instead of the name "root", how about "sim" or the like?  The command line programs are for simulation, and the utreexo package will be built in to wallet / node software, with probably no user-facing executables in this repo, so "sim" might be a more descriptive name.

kcalvinalvin:
I tried implementing the standard library flags but had to rewrite code because of the flags ttlfn and schedFileName. The standard library doesn't support subcommands for different commands. To implement the standard library, there would have to be a separate subcommand written for each flag. I guess it's fine if there's no more plans to add more commands within simulation but I think it'll start becoming messy real quick if there is going to be.

Changed the name of root.go to sim.go.

adiabat:
The way I've done flags for various commands before is to have a more generic `flag.FlagSet` with variables like `inputFile` and `outputFile`, and then the different commands can use them for different purposes.

This is how I've done it in other programs:
```
type simSession struct {
	command string
	inFileName *string // input filename 
	outFileName *string // output filename
}

func (s *simSession) setFlags(fset *flag.FlagSet) {
	s.inFileName = fset.String("i", "", "input file name")
	s.outFileName = fset.String("o", "", "output file name")
}

func main() {
	if len(os.Args) < 2 {
 		return
	}

	fSet := flag.NewFlagSet(os.Args[0], flag.ExitOnError)

	var session simSession
	session.setFlags(fSet)
	session.command = os.Args[1]
	fSet.Parse(os.Args[2:])
	// do stuff
	return
}
```

maybe there's a better way to do this?  It's pretty simple and only uses standard library.  

That said, standard library flag parsing is also weird.  Why does it want `os.Args[0]`? Also a bit weird that you have to use pointers for everything in the session struct, but that's what `flag.FlagSet` gives you.  And that `fset.Parse` comes at the end and causes the pointers in `session` to point to the right thing.  Does seem to work even if the sequence is initially confusing.  But it doesn't have an `init()` function which I also don't really like :)

Cobra does seem OK and set up for more complex flag parsing, I'll look at it more today.  So far this has very limited flag parsing needs so I think we can stick with standard library for now.

kcalvinalvin:
Removed Cobra and implemented the standard flag library.

I think this is the cleanest you can get with the standard lib.

adiabat:
Cool, thanks, looks good.
(I might tweak some things later but nothing major, some of the parse stuff; `os.Args[2:3]` seems scary..)

PR #16, opened 2019-10-12
[WIP] CLI and CMD refactor

GregTheGreek:
This pr introduces a CLI and re-structures the project into more useable components.

It also has chainsafe imports because of some local development i was tinkering with..


readme.md
```diff
@@ -35,13 +35,13 @@ $ bitcoin-cli stop
 * get utreexo code
 
 ```
-$ go get github.com/mit-dci/utreexo
+$ go get github.com/chainsafe/utreexo
```
kcalvinalvin: Should change all of these back to mit-dci for a PR

GregTheGreek: addressed

node/main.go
```diff
@@ -0,0 +1,15 @@
+package node
+
+import (
+	log "github.com/ChainSafe/log15"
```
kcalvinalvin: Wouldn't it be better if the changes were pushed upstream and that was imported?

kcalvinalvin:
Haven't looked through everything but seems like the main difference is that you've already added the ability to add a config file along with logging.

node/main.go
GregTheGreek: Not sure im following you? 

This is a custom package, just a simple logger.

kcalvinalvin: hmm never mind, the original repo that you guys forked off of seems abandoned.

GregTheGreek: ohhh, yeah sorry if that wasn't clear. We also updated it a bit.

PR #19, opened 2019-10-22
Fix minor error in sim.go

kcalvinalvin:
Fixes the problem with program not running unless you give it 3 arguments. Also fixes the [2:3] array issue.

Sorry for the simple dumb mistake. This fixes them.


adiabat:
No problem, thanks!  I was starting to fix this part, but now don't have to :)

PR #21, opened 2019-10-24
Replace blake256 with sha256

kcalvinalvin:
Simple replacement of hash algorithms. I think it's slower but oh well.


adiabat:
Sure, can use sha256.

I wonder if sha512 is faster; it's used a bunch in bitcoin-core so maybe people would be OK with that too.

PR #25, opened 2019-11-01
Add stop and resume feature for parseblock

kcalvinalvin:
Adds stop and resume for blockparser.

Behavior:

1) Builds `offsetfile` && `currentoffsetheightfile` first. "offsetfile" stores the 8 byte info of `.dat file num` and `offset`. `currentoffsetheightfile` is the tip from the offsetfile since there are is no tipnum saved to it. This is a cheaper way of doing that. If offsetfile is found, that file is used for syncing. If not found, it will build one. If there was more blocks synced by Bitcoin Core afterwards, simply delete the `offsetfile` and `currentoffsetheightfile` and it will build a new one based on the new .dat files given. Haven't tried with the full chain yet but it shouldn't take more than a minute at most.
If SIGINT, SIGTERM, or SIGQUIT is passed while building the two files, it will delete them since they are incomplete. 

2) Blocks are read off the offsetfile and writes them like the old software would.

### DIFF
--- a/cmd/blockparser/blockparser.go
+++ b/cmd/blockparser/blockparser.go
@@ -1,8 +1,13 @@
 package blockparser
 
 import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/binary"
+	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"sync"
 
 	"github.com/mit-dci/lit/btcutil/chaincfg/chainhash"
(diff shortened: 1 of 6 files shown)


kcalvinalvin:
There's some more things I need to work out. I'll force push them soonish

kcalvinalvin:
Ready to be merged now.

As talked about on IRC, the 80 byte headers are read off the .dat files. Each block's offset and .dat file is written to "offsetfile" and the last synced height for offsetfile is saved to "currentoffsetheight".

To update the offsetfile, just delete and restart and simcmd will build it again with the current available .dat files in the directory.

Generated .txos file is identical to how it is being done on master right now.

kcalvinalvin:
Ran some benchmarks today. I've only tested up til blk00007.dat on testnet but the new way seems to be 3% faster so there is that.

adiabat:
tested, ACK, works great.

The only thing I'm worried about is the state of the levelDB batches when `stopParse()` does `os.Exit(1)`

(also should it exit 1 instead of 0?  Does that even do anything? not sure the convention)

We can look at changing things later, definitely merge-able now.

adiabat:
oh and the `RawHeaderData` struct, actually in this case it probably makes no difference, but I usually try to have bigger things in the beginning then smaller things at the end, like:
```	CurrentHeaderHash [32]byte
	Prevhash [32]byte
	FileNum [4]byte
	Offset [4]byte```

But I think the FileNum and Offset stick together into one 8-byte thing here so should work either way.

**Comment by adiabat** [comment]
the `defer lvdb.Close()` is the only thing that worries me.  Maybe it works fine but maybe levelDB messes up sometimes if it doesn't get closed.  One way to close it nicely would be to change the main loop
```
for ; tipnum != currentOffsetHeight; tipnum++ {
``` 
to something that has sig in it as well, like `tipnum ! = currentOffsetHeight && keepGoing` where `keepGoing` is a bool that `stopParse()` modifies or something like that.

Then it still does the `batchwg.Wait()` at the end and closes the db.

if there are any issues with the ttldb we should change this first since that would be a likely source of problems.

adiabat:
Also `getTipNum()` is overkill -- the last line of testnet.txos will always be h: 00000, and if it's not things are screwed up enough that it's fine to give an error and quit.  So you can just directly read the last line.

"h: 1234567" is 10 bytes long, and we can go to 16 bytes (no line should be 6 bytes long), so just `f.Seek(16, 2)` then read the last line

Alternatively we can zero-pad the h: 0000 line so that it's a fixed length.  Actually I kindof like that better, then you know the exact place to seek and read.

... or I guess can just leave all this until we switch to binary

PR #28, opened 2019-11-19
 Fixed go-lint, added testnet flag, made stopParse wait for ttldb

kcalvinalvin:
Now flag -testnet=true will handle things. No need to do ttlfn=ttl.testnet.txos now.
RawHeaderData Struct is also organized. 
It also added wg to stopParse() but not sure if that does anything.

### DIFF
--- a/cmd/blockparser/blockparser.go
+++ b/cmd/blockparser/blockparser.go
@@ -31,10 +31,10 @@ make a txo text file with the ttl times on each line
 //CurrentHeaderHash is the 80byte header double hashed
 //Prevhash is the 32 byte previous header included in the 80byte header.
 type RawHeaderData struct {
-	FileNum [4]byte
-	Offset [4]byte
 	CurrentHeaderHash [32]byte
-	Prevhash [32]byte
+	Prevhash          [32]byte
+	FileNum           [4]byte
+	Offset            [4]byte
 }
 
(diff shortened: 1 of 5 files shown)


kcalvinalvin:
Changed how the main loop in blockparser.go works based on the tuesday call.

Now, when given SIGINT, SIGTERM, or SIGQUIT, it will wait until the ttldb closes. There's a bit of a delay now depending on where in the loop it is because of this.

cmd/blockparser/blockparser.go
```diff
@@ -114,9 +132,17 @@ func Parser(sig chan bool) error {
 
 	fmt.Println("Building the .txos file...")
 	fmt.Println("Starting from block:", tipnum)
+
+	//Receive signal from stopParse so the for loop can break
+	var stop bool
+	go func() {
```
adiabat: I guess this works!  But it seems weird because stop is being affected in its own goroutine.

Instead, how about this at the bottom of the loop (line 159）
```
		select {
		case stop = <-stopGoing:
		default:
		}
```

this might actually do the exact same thing (they both seem to work) but this feels a lot safer / easier to understand for me as there are fewer concurrent routines happening.

adiabat:
Looks good & works, just 1 requested code change

PR #29, opened 2019-11-23
Add resume to txottl

kcalvinalvin:
I worked off the pull #28 for this one though. Can rebase onto master if #28 isn't good.

### DIFF
--- a/cmd/txottl/txottlparse.go
+++ b/cmd/txottl/txottlparse.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/mit-dci/utreexo/cmd/blockparser"
 	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/opt"
 )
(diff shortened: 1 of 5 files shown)


cmd/txottl/txottlparse.go
```diff
@@ -230,23 +285,60 @@ func ReadTTLdb(sig chan bool) error {
...
+			//Only write where we left off after a block has been finished
+			ttlfilesync.WriteAt(U32tB(uint32(position))[:], 0)
 
 			// start next block
 			// wipe all block txs
 			blocktxs = []*txotx{new(txotx)}
 
+			//Check if stopSig is no longer false
+			//stop = true makes the loop exit
+			if stopSig == true {
```
adiabat: same idea here; replacing this with
```
	        select {
		case stop = <-stopGoing:
		default:
		}
```
lets you get rid of the goroutine on line 188.  It also would let you get rid of the whole `stopTxottl` routine if it were 

```
	        select {
		case stop = <-sig:
		default:
		}
```

which seems simpler as then there's 1 channel instead of 3

adiabat: Wait, not 100% sure on that last one.  In `stopTxottl` the bool coming out of `sig` is just dropped, so the thing I wrote would only work when `sig` has a true in it, but not a false.  

So could still work, just that's a difference to be aware of.

kcalvinalvin: Implemented requested changes for both #28 and #29

adiabat:
Looks good.  I might change a little bit in parseblock but I guess it's all changing pretty substantially soon so should be fine.

PR #32, opened 2019-12-02
Refactor and removed testnet.txos

kcalvinalvin:
This works **but** is around 10~15 percent slower to generate the ttl.testnet.txos file. I think it's because GetRawBlockFromFile() is calling Deserialize() https://github.com/mit-dci/utreexo/blob/c9d6035a87db07e5ba2ff07fa6c14a317bcfb24b/cmd/blockparser/blockparser.go#L427-L428

twice as much than it did before (so we're losing a lot of benefits gained here). This could be optimized I think since we don't need everything that Deserialize() gives. Would also make parseblock a **lot** faster too.


adiabat:
OK - slower but as per conversation here's how to make it (hopefully) way faster:

`BuildProofs()` does not need TTL data.  So `BuildProofs()` can also do the job of generating the TTL db, since it's gotta do deserialize on all the blocks anyway.  Should be 2 birds / 1 stone.  

Then `RunIBD()` won't have a text file to access, so it will need to run deserialize on the blocks, and the lookup ttl values in the DB itself instead of having everything in a text file.  So `RunIBD()` might be a little slower but we're saving an entire pass through the blockchain since txottl is merged into `BuildProofs()`

PR #36, opened 2019-12-17
Remove txottl and merge blockparser with genproofs

kcalvinalvin:
I left histogram stuff alone in simcmd but everything is commented out in histogram.go and doesn’t work.

Everything else works great.


kcalvinalvin:
ahhh geez I benchmarked it and it’s 70% slower than reading it off a ttl file :/

GetBlockFromFile() is suuuuper slow :(

kcalvinalvin:
Ugh ok I did some benchmarks and this new way is around 20~25 percent slower.

https://github.com/mit-dci/utreexo/blob/b5f1f20d25b8cc273fdec1626361c3c65bc7e5ba/cmd/utils/utils.go#L66-L68
This function is incredibly slow.. :(

kcalvinalvin:
Removed some fmt.Print stuff that I had in utreexo/utreexo for testing.

kcalvinalvin:
Merged blockparser.go with genproofs.go as discussed. 

Benchmarking shows the performance to be about the same as the old method. I haven't done concurrency things yet so it definitely can be faster.

adiabat:
tested ACK / LGTM

as discussed - stop / resume isn't working but should be straightforward for genproofs:

https://github.com/kcalvinalvin/utreexo/blob/9fb32d2357117351eda7b5d5b9854cb06cf1d497/cmd/ibdsim/genproofs.go#L169-L172

switch that around.  Also, for speeding it up, I think a good way to do it is have `GetRawBlockFromFile` happen concurrently with everything else and get pushed into a buffered channel.  So have a "blockReader" function:

```
func blockreader(blockchan chan []*wire.MsgTx, height int, offsetfile *os.File) {
	for {
		block, err := simutil.GetRawBlockFromFile(height, offsetfile)
		if err != nil {
			panic(err)
		}
		blockchan <- block
		height++
	}
}
```

and then blockchan is read from in the main loop.  As `GetRawBlockFromFile` is faster than the `writeProofs` stuff going on, it should get ahead.  Maybe a buffer size of 10 or 20 or something.

adiabat:
Also I can't get it to make the nice quotes from a code selection.  Oh well I'll figure it out again at some point :)

PR #38, opened 2019-12-22
Added goroutines for reading blocks off the .dat files

kcalvinalvin:
Goroutine added for ReadRawBlockFromFile() with Blockreader() inside utils.go

Benchmark shows around 12% improvement when run 
```./simcmd genproofs -testnet=1 && ./simcmd ibdsim -testnet=1```

compared to the old way.

### DIFF
--- a/cmd/ibdsim/genproofs.go
+++ b/cmd/ibdsim/genproofs.go
@@ -16,17 +16,20 @@ import (
 // build the bridge node / proofs
 func BuildProofs(isTestnet bool, ttldb string, offsetfile string, sig chan bool) error {
 
-	//Channel to alert the main loop to break
+	// Channel to alert the main loop to break
 	stopGoing := make(chan bool, 1)
 
-	//Channel to alert stopTxottl it's ok to exit
+	// Channel to alert stopTxottl it's ok to exit
 	done := make(chan bool, 1)
 
-	//Channel to alert stopBuildProofs() that buildOffsetFile() has been finished
+	// Channel to alert stopBuildProofs() that buildOffsetFile() has been finished
 	offsetfinished := make(chan bool, 1)
 
+	// Channel for stopBuildProofs() to wait
+	finish := make(chan bool, 1)
+
 	// Handle user interruptions
-	go stopBuildProofs(sig, offsetfinished, stopGoing, done, offsetfile)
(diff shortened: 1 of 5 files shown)


adiabat:
Looks good, seems to work-
In terms of speed it has removed the block reading / deserialization bottleneck, but it's still pretty slow because of all the other single-threaded stuff.  I'm running it on an 8 core machine and it's only at 150% cpu or so.

https://github.com/mit-dci/utreexo/blob/c2b7373d1a6d049885d7a601a8acf15eec36ed3e/cmd/ibdsim/genproofs.go#L252

this part is just a sanity check; all it does is fail / end the program if it's not OK, but it doesn't affect `newForest`.  So we can take that out; you can see how much that speeds it up.

The real bottleneck though is `Modify`, which... I'm working on :)
Modify is:

`f.reMap` -- ~never happens (20 times throughout the whole process) so whatever

`f.removev3(dels)` that's the slow one which will be multithreaded as soon as I merge the undo branch.

`f.addv2(adds)`  also completely sequential right now, haven't even looked at that one.  The hashing step can definitely be made multithreaded but it's significantly more complex than the add method is now, which is super simple. (Though a multi threaded add is definitely simpler than the multithreaded remove).

kcalvinalvin:
> In terms of speed it has removed the block reading / deserialization bottleneck, but it's still pretty slow because of all the other single-threaded stuff. I'm running it on an 8 core machine and it's only at 150% cpu or so.

This is something that I've been looking at too. On my desktop with 6 cores, cpu load is tiny. But on my laptop with 2 cores, cpu is the bottleneck. I think Bitcoin Core would have said it's good enough and moved on but not sure what the goal of this repo is.
 
> https://github.com/mit-dci/utreexo/blob/c2b7373d1a6d049885d7a601a8acf15eec36ed3e/cmd/ibdsim/genproofs.go#L252
> 
> this part is just a sanity check; all it does is fail / end the program if it's not OK, but it doesn't affect `newForest`. So we can take that out; you can see how much that speeds it up.

I'll give it a go.

> `f.addv2(adds)` also completely sequential right now, haven't even looked at that one. The hashing step can definitely be made multithreaded but it's significantly more complex than the add method is now, which is super simple. (Though a multi threaded add is definitely simpler than the multithreaded remove).

Been kinda scared to touch the library parts to avoid all git jumbled mess. But if you aren't looking at addv2, I'll look at it.

Issue #44, opened 2020-01-02
UTXOs have negative duration


kcalvinalvin:
ttldb returns 0 if the tx wasn't found and all the utxos end up having negative duration since 

```deathheight - height```

is applied to all txos. So the ones that get 0 returned have negative duration.

In the older commits with the .txos files, the 0 heights had "s" appended to them. What should we do for these now?

kcalvinalvin:
ok so "s" are given 1,000,000 (1 << 20).

Should this be implemented?

adiabat:
I think this is OK?  A little ugly but have some large value (could be 1<<30) for txos which stay unspend forever.

PR #45, opened 2020-01-03
Added separate arg named "clair-ibdsim" to do ibdsim with clair

kcalvinalvin:
Reworked clair into master. Works again and it's about the same as without clair?

Also changed some things with genLeafTXO() and is a bit faster now.


kcalvinalvin:
Fixes #44

PR #47, opened 2020-01-09
updated undo branch

adiabat:
forest and pollard use remove transform 2, getting rid of the old remove transform


kcalvinalvin:
What are the major changes made to the Forest and Pollard lib?

PR #50, opened 2020-01-13
Disk resume and remove opening files in Lib

kcalvinalvin:
Added functions to resume forest when using disk.
Cleaned forestdata.go and forest.go functions so they don't open up
files. Moved all the file opening functions to cmd/

So this works but is *painfully* slow.


kcalvinalvin:
hmm ok I could rebase onto the new master branch. Should I do that or do you have more things to push?

PR #59, opened 2020-01-21
add rev reader

tnakagawa:
I created a code to read revNNNNN.dat of Bitcoin.

'rev.go' is the main code.
'rev_test.go' is the test code.
'rev.md' is a description of the file format.

Confirmed to read all Mainnet rev files.

__Problem__

- I refer to the design of "archive / zip".
Therefore, I use 'io.ReaderAt'. Is 'io.Reader' better?

- Add test code to read files.

- rev.go does not use anything other than the standard library.
Is it better to use lit library etc?
However, rev_test.go uses the lit library.


kcalvinalvin:
Some general structural changes that I think is good:

1. Move code to ```cmd/``` directory.
The rev*.dat files are specific to Bitcoin Core so it'd be better fitting in the ```cmd/``` directory. A new ```rev``` directory inside ```cmd/simutil/``` would be good. 

2. Since the rev*.dat file's purpose is for un-doing blocks, realistically the code will only need to call one rev*.dat file (and would only need one rev block). The ```init()``` function is currently used to grab all rev blocks in a rev*.dat file but it's not how it's going to be used.
Suggestion is to get rid of the for loop on line 78 of rev.go and have ```init()``` take in an offset as an argument. Then you could get rid of variable ```pos``` and just use os.Read(). Then it would return the rev data and there would need to be more code to handle that.

3. This is more of an additional feature but there would need to be a way to get the offset and rev*.dat file name for a certain block # from a rev*.dat file. Right now, leveldb stores everything in Bitcoin Core but we'd need to generate our own index. And since the blocks in rev files aren't guaranteed to be in order (or match the order of the blocks in the blk*.dat files), there would need to be checks against that as well.

utreexo/rev.go
```diff
@@ -0,0 +1,352 @@
...
+	}
+	rr := new(RevReader)
+	err := rr.init(r, size)
+	if err != nil {
+		return nil, err
+	}
+	return rr, nil
+}
+
+func (rr *RevReader) init(r io.ReaderAt, size int64) error {
```
kcalvinalvin: I think it would be better to:

1.  Change from a method to a function that accepts a variable of type RevReader as an argument.

2. Remove size as an argument (since you'll only be reading one block).

3. Add offset (where the block is inside a file) as an argument.

4. Add filename (which rev*.dat file the block is in) as an argument.

utreexo/rev.go
```diff
@@ -0,0 +1,352 @@
...
+
+// CTxInUndo ...
+type CTxInUndo struct {
+	Height uint64
+	Script []byte
+	Amount uint64
+}
+
+// OpenRevReader ...
+func OpenRevReader(name string) (*RevReader, error) {
```
kcalvinalvin: Probably won't need this function. Could delete it.

utreexo/rev.go
```diff
@@ -0,0 +1,352 @@
...
+	err = r.init(f, fi.Size())
+	if err != nil {
+		f.Close()
+		return nil, err
+	}
+	return r, nil
+}
+
+// NewRevReader ...
+func NewRevReader(r io.ReaderAt, size int64) (*RevReader, error) {
```
kcalvinalvin: Probably won't need this function either. Could delete this function as well.

tnakagawa:
@kcalvinalvin 
Thank you for your comment.

> Some general structural changes that I think is good:
> 
> 1. Move code to `cmd/` directory.
>    The rev*.dat files are specific to Bitcoin Core so it'd be better fitting in the `cmd/` directory. A new `rev` directory inside `cmd/simutil/` would be good.

You're right.
This code does not need to be in this repository.
This PR is closed because it seems better to open it as my personal library.

> 2. Since the rev*.dat file's purpose is for un-doing blocks, realistically the code will only need to call one rev*.dat file (and would only need one rev block). The `init()` function is currently used to grab all rev blocks in a rev*.dat file but it's not how it's going to be used.
>    Suggestion is to get rid of the for loop on line 78 of rev.go and have `init()` take in an offset as an argument. Then you could get rid of variable `pos` and just use os.Read(). Then it would return the rev data and there would need to be more code to handle that.

I think it is good to make the code that analyzes one data a public function.

But, I think it is a difference in use.
This code converts all the data in the rev * .dat files into golang objects, not to extract one data.

> 3. This is more of an additional feature but there would need to be a way to get the offset and rev*.dat file name for a certain block # from a rev*.dat file. Right now, leveldb stores everything in Bitcoin Core but we'd need to generate our own index. And since the blocks in rev files aren't guaranteed to be in order (or match the order of the blocks in the blk*.dat files), there would need to be checks against that as well.

I understand the content.
However, it is outside the scope of this code.

kcalvinalvin:
I've just tested the code and it seems to work well.

> You're right.
> This code does not need to be in this repository.
> This PR is closed because it seems better to open it as my personal library.

You could force push to your branch. Whatever is better for you.

> But, I think it is a difference in use.
> This code converts all the data in the rev * .dat files into golang objects, not to extract one data.

Yeah I see that and it does seem pretty fast to read one rev*.dat file so probably fine to leave it like this for now.

> I understand the content.
> However, it is outside the scope of this code.

Sure. Something to keep in mind for the future

utreexo/rev.go
```diff
@@ -0,0 +1,352 @@
...
+)
+
+// RevReader ...
+type RevReader struct {
+	Records []*BlockRecord
+}
+
+// BlockRecord ...
+type BlockRecord struct {
+	Magic uint32
```
kcalvinalvin: Magic should be ```[4]byte``` since it's just for checking. The rest of the repo has it as ```[4]byte``` as well.

utreexo/rev.go
```diff
@@ -0,0 +1,352 @@
...
+	pos := int64(0)
+	for size > pos {
+		record := new(BlockRecord)
+		buf := make([]byte, 4)
+		_, err := r.ReadAt(buf, pos)
+		if err != nil {
+			return err
+		}
+		record.Magic = binary.LittleEndian.Uint32(buf)
+		if record.Magic == 0 {
```
kcalvinalvin: Here, since we know the magic bytes of testnet and mainnet, you could hold a test to see if it matches.

utreexo/rev.go
tnakagawa: I also thought byte[4] was good.
However, Lit uses uint32.

https://github.com/mit-dci/lit/blob/master/wire/protocol.go#L114

kcalvinalvin: hmmm ok. @adiabat ?

I guess you can leave it alone for now

tnakagawa:
Update

- []byte -> io.Reader
- The UndoReadFromReader function is that read one undo.

utreexo/rev.go
tnakagawa: The current network is limited, but there are no restrictions for Magic (mainnet, testnet, regtestnet, signet, etc ...), so I decided to remove this check.

adiabat:
Hi!  どーも、レビュー遅くてすみません

Looks good - one thing to change that's pretty simple is to use `==` instead of reflect, I think it's easier simpler.  Then OK to merge and we can put the rev data into leaves.

tnakagawa:
@adiabat 
Is it ref_test.go reflect.DeepEqual?
Golang slices cannot use ==.

adiabat:
@tnakagawa right, slices can't use `==`, but in this case, we don't need to use slices for hashes with a known length.  We can use `[32]byte` which can use `==`.  In cases where we need a slice, such as `copy`, we can use `hash[:]` which converts the array into a slice.  It's easy to convert from array to slice, but hard to go from slice to array, so we should keep the hashes as arrays.

tnakagawa:
I am reluctant, but I fixed it.
When comparing sequences in Golang, it is common practice to use reflect.DeepEqual.

adiabat:
If you don't want to use `==` operator, you can use `bytes.Equal()` instead; I've used that in many programs but not so far in utreexo as so far everything is fixed length.

I think reflect is generally not used in go.  For example at https://golangbot.com/reflection/ -- "It should be avoided wherever possible and should be used only when absolutely necessary."

also, according to these benchmarks:
https://gist.github.com/fishy/bca72d0d65ba35e9ab6b50f66f9fc42e
using reflect is thousands of times slower than `bytes.Equal()`

tnakagawa:
understood.
== changed to bytes.Equal

tnakagawa:
> I think reflect is generally not used in go. For example Rob Pike, one of the main authors of go, says 
> "It's a powerful tool that should be used with care and avoided unless strictly necessary."(https://blog.golang.org/laws-of-reflection)

What he is talking about is the ability to access objects with reflect.
This should be used with care, as it can destroy the object itself.
Does reflect.DeepEqual also apply?
I do not think so.
I feel like you're over-expanding.

adiabat:
OK, we don't have to get in to use of reflect; I'm just generally wary of it.  Regardless of that, it's thousands of times slower for short byte slice comparisons than `bytes.Equal`, so in this case can we use `bytes.Equal` in rev_test.go?

tnakagawa:
Updated code.

I am not convinced not to use reflect.
It's clear from the code that I'm not using reflcet in dangerous ways.
Also, I did not use bytes.Equal early on.
This is because bytes.Equal is more suitable for use than reflect.DeepEqual.
You disagree with the proposed PopCount speedup.
It just seems like you've decided at your own discretion.
Is this github open source? Is it your workspace?
Please define the code rules and CI as soon as possible.

kcalvinalvin:
Open Source does not mean that every one of your PRs have to be merged.

It is up to the author of the PR to *convince* the maintainer that the changes you are proposing is worth:

1. Reviewing
2. Maintaining

It’s the same with *all* Open Source projects. Many projects outright ignore PRs if the maintainer deems it to be not to be worth it.

If you disagree with the direction of the project or the changes that are being made, it is general knowledge with Open Source that you should make a different repository and continue a separate project there.

tnakagawa:
My official language is Japanese and I am not good at English.
I basically use Google Translate, so I don't know if it makes sense.

I am not saying that my PR will not be merged.

Whether or not a library is used should be described as a minimum rule.

Previously, the test code in master was in error.

The evaluation criteria also look vague, so it looks like they are just saying something convenient.

If you want to conform to Golang standards, you should also set lnt such as formatter.

In fact, I am different from my policy, but amend it according to your policy.

Shouldn't I give free opinion here?

Do you want to do it in your repository?

tnakagawa:
I'm sorry, I said too much.
Put it in its own repository.

PR #66, opened 2020-01-27
Add Pollard resume functionality.

kcalvinalvin:
Only stores the tops in pollard at the moment. Does resume though.


adiabat:
Cool, this is enough for a while actually.  Keeping partial forest data in RAM-only makes sense.  Like how bitcoin would keep the mempool only in memory... though now the "mem"pool is flushed to and restored from disk, it works fine without doing that.

Issue #68, opened 2020-01-28
Does resize() need to be called on diskForestData?


kcalvinalvin:
In forest.go, several places all resize() on the diskForestData. But in the case of https://github.com/mit-dci/utreexo/blob/912eb23a04aabc3ccac02aecab162c1ddf758e25/utreexo/forest.go#L84

if you resume with diskForestData and that line isn't commented out, it'll wipe the whole forestfile.dat and replace with ```0```s.

Not sure if they're needed in some places but intuition says it shouldn't be called for diskForest

adiabat:
It maybe doesn't!  I'm not 100% sure; if you try to write to a specific offset in a file, and the file isn't big enough, does it auto-extend the file to the required size (and fill with 0s) or does it give an error?

However, I think a better way to deal with this is to have two separate functions, `NewForest(forestFile *os.File) *Forest` which gives you a brand new forest and creates / wipes out whatever's at that file pointer, and `LoadForest(forestFile *os.File) (*Forest, error)` which loads the forest from an existing file, re-creates the positionMap, gives errors if something's wrong with it, etc.

PR #69, opened 2020-01-30
utreexo/forest.go: Changes to RestoreForest()

kcalvinalvin:
Changes RestoreForest() from a method to a function
No longer relies on calling newForest() in cmd/

Addresses the comment by @adiabat on #68


adiabat:
ACK

PR #72, opened 2020-01-31
Fix auntable() comment

kcalvinalvin:
So there is a pretty big typo where you call the function auntable() as auntOp. Not sure if you like to have a bunch of commits for one line changes so this could be fixed in the next big(er) thing too.


adiabat:
heh yeah that looks obviously wrong :)

PR #78, opened 2020-02-07
cmd/genproofs.go: Fix bug where user interruptions didn't quit

kcalvinalvin:
During genproofs, when offsetfile was being built, user interruptions
were ignored. Fixes this.


kcalvinalvin:
Also switched some ```return(err)``` to ```panic(err)``` since they were impossible to debug since I didn't know where it came from

PR #82, opened 2020-02-16
Change the link for irc logs to kanzure's website

kcalvinalvin:
Kanzure's setup has live updates while my logs only update every midnight. Far better logs.


adiabat:
Cool thanks!

Issue #79, opened 2020-02-10
Goals for upcoming May Utreexo Workshop


kcalvinalvin:
This is a list of goals for the upcoming May Utreexo workshop. Having a Bridge Node and a well documented code would help with newcomers to the project.

I think the core libraries are stable enough to go make a bridge node now. I thought of some things the bridge node should do. Please correct/add to the list as necessary.

# Nodes

##### B. Serve Classic Bitcoin Nodes as well as Utreexo Compact State Nodes as stated in the paper.

1. Keep blk*.dat files and rev*.dat files like Bitcoin Core.
2. Keep Proofs and the Compact state separately in a different directory.

##### C. Contain a mempool.

1. Generate Proof for TX from a Classic Bitcoin Node.
2. Serve both the Classic Bitcoin Node as well as Utreexo Compact State Nodes.

##### A. Index the locations of a block in a blk*.dat file. We have three options:

1. Use our offsetfile method for indexing. It would just do them on the fly instead of reading from a given local blk*.dat file.
2. Use Leveldb. 
3. Receive the blocks sequentially. (Would be slower)

Open to thoughts here.

##### A. Set up Bridge Node(s) for Utreexo Compact State Nodes to query.

Make a Bridge Node with MIT's resources?

##### B. Set up Utreexo Compact State Node(s)

Should be easy.

## 3. Utreexo API

A list of calls to provide so anyone can fork and start using the library. Would also help in seeing how Utreexo could fit into Bitcoin Core. This is an attempt to list what there should be. Please suggest additions/subtractions. Does fit into the Bridge Node and the Utreexo Compact Node construction

# Documentation

Make godoc like btcd.

## 1. README.md and doc.go for every directory

There are old comments that need cleanup and there are also functions with outdated/no comments.

## 3. A medium article or something similar would be nice.

adiabat:
Cool - generally agree; some comments:

1. Make Standalone Bridge Node - this is pretty much done.  Not too much left here, and we don't need it to support normal bitcoin full nodes, or even do any validation itself.  Initially it can be just a server that feeds data do CSNs (compact state nodes)

I also think that mempool can come later, after we get block proofs & validation on both sides.  There are some optimizations & tricky parts about dealing with mempool that we can leave till later.  Blocks-only mode is a thing, and gives plenty of functionality for proving the concept.

2. CSN - this part is the harder one because of the need to do block validation, without a database.  The function we need is

[...]

PR #86, opened 2020-02-23
Refactor utreexo/cmd/

kcalvinalvin:
Everything in ```utreexo/cmd/``` is refactored.

While reviewing code to do the TODOs mentioned in #79, I noticed some things to clean up.

### This refactor is needed for the goals mentioned in #79 as it separates

1. Bridge node and CSN (Compact State Node)
2. Chain resume functionality from both bridgenode and csn (placed in ```chain.go``` and ```chainio.go``` in both ```bridgenode/``` and ```csn/```

#### Changed things

1. All functionality for store and restore to disk is now placed in ```chain.go``` and ```chainio.go``` in both ```bridgenode/``` and ```csn/```

2. Various bugs involving offsetfile.go was fixed. To add new blocks from ```blk*.dat``` files, all the user has to do is delete the ```offsetdata``` directory and re-run ```genproofs``` to rebuild the offset index. Both ```genproofs``` and ```ibdsim``` will resume where they left off and sync the new blocks.

3. ```main.go``` file in directory ```cmd/``` is the main file now. Also
removed some unused variables.

4. Separated all ttl related things to its own directory.

5. Updated ```readme.md``` in ```utreexo/cmd/``` and in ```utreexo/```

6. Removed all names related or implying a simulator.

7. Various comment cleanups in all files


somsakpanom:
รับBitcoเหริญ สนับสนุนเพื่อที่จะสามารถเอาไปแปลงแลกเปลี่ยนเป็นเงินสดใช้จ่ายได้

adiabat:
This is great -- I like how it's organized a lot better.  There are a couple little things I want to fix / change, but I think it's better to merge it and tweak stuff later so we can base stuff off of this code reorg.

adiabat:
post-merge comments:

we should call this something other than "cmd" :)

it could be ... utreexo, and then the library is libutreexo?  Or utreexo stays the library, and the node binaries are something else?  There's a lot of logic and things going on that's outside of the library at this point, so it needs it's own name; it's still "simulating" IBD, but we're getting past the point of being a simulation; once tx validation is in, it's basically IBD.

Simutils -> util, but we should call that something else as well.  It's utils for the cmd.

utreexo -> libutreexo, cmd -> utreexo, util -> utilutreexo? utrils? 
or cmd can be "unode" or "csn" or something.

Anyway it's basically just naming things which is one of the 2 or 3 hard problems in CS :)

kcalvinalvin:
I... must admit I'm terrible at naming things. If you feel like something is better, do change them!

PR #88, opened 2020-02-28
Add Regtest support

ysangkok:
But caution, regtest is called TestNet in Lit.
(and testnet is called TestNet3)

For user-facing text, we use the 'regtest' name
to lessen the confusion.

### DIFF
--- a/cmd/bridgenode/chain.go
+++ b/cmd/bridgenode/chain.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/mit-dci/lit/wire"
 	"github.com/mit-dci/utreexo/cmd/util"
 	"github.com/mit-dci/utreexo/utreexo"
 )
@@ -11,7 +12,7 @@ import (
 // initBridgeNodeState attempts to load and initialize the chain state from the disk.
 // If a chain state is not present, chain is initialized to the genesis
 // returns forest, height, lastIndexOffsetHeight, pOffset and error
-func initBridgeNodeState(isTestnet bool, offsetFinished chan bool) (
+func initBridgeNodeState(net wire.BitcoinNet, offsetFinished chan bool) (
 	*utreexo.Forest, int32, int32, int32, error) {
 
 	var offsetInitialized, forestInitialized bool
@@ -33,7 +34,7 @@ func initBridgeNodeState(isTestnet bool, offsetFinished chan bool) (
 	} else {
 		var err error
 		fmt.Println("Offsetfile not present. Indexing offset for blocks blk*.dat files...")
-		lastIndexOffsetHeight, err = createOffsetData(isTestnet, offsetFinished)
+		lastIndexOffsetHeight, err = createOffsetData(net, offsetFinished)
 		if err != nil {
 			return nil, 0, 0, 0, err
 		}
--- a/cmd/util/utils.go
+++ b/cmd/util/utils.go
@@ -10,7 +10,7 @@ import (
 )
 
 // Hash is just [32]byte
-var mainnetGenHash = Hash{
+var mainNetGenHash = Hash{
 	0x6f, 0xe2, 0x8c, 0x0a, 0xb6, 0xf1, 0xb3, 0x72,
 	0xc1, 0xa6, 0xa2, 0x46, 0xae, 0x63, 0xf7, 0x4f,
 	0x93, 0x1e, 0x83, 0x65, 0xe1, 0x5a, 0x08, 0x9c,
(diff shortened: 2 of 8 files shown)


ysangkok:
Wait, I just found an issue with the command flag parsing.

cmd/util/utils.go
```diff
@@ -24,39 +24,51 @@ var testNet3GenHash = Hash{
...
+func GenHashForNet(net wire.BitcoinNet) Hash {
+	switch net {
+		case wire.TestNet3:
+			return testNet3GenHash
+		case wire.MainNet:
+			return mainNetGenHash
+		case wire.TestNet: // yes, this is regtest
+			return regTestGenHash
+	}
+	panic("net not supported by GenHashForNet")
```
kcalvinalvin: I think returning and error here is better. In some parts of the code, you'll see functions with ```panic(err)``` instead of ```return err``` and this was mostly for debugging since with panics, you get the line # of where it errored out.

I think something like
``` func GenHashForNet(net wire.BitcoinNet) (Hash, error)```
and
``` return nil, fmt.Errorf("net not supported\n")```
is good

File: cmd/util/utils.go [hunk 2]
```diff
@@ -24,39 +24,51 @@ var testNet3GenHash = Hash{
...
+	f, err := os.Open("blk00000.dat")
+	if err != nil {
+		panic(err)
+	}
+	defer f.Close()
+	var magicbytes [4]byte
+	f.Read(magicbytes[:])
+
+	bytesToMatch := make([]byte, 4)
+	// Can't use U32tB because we need little-endian
```
kcalvinalvin: In ```cmd/util/utils.go```, there is a function that does little-endian to u32.

I think it would be better if a ```U32tLB()``` was added and called here

kcalvinalvin:
A side note but the reason Regtest wasn't really implemented was because Utreexo testing is mostly on the transactions and there aren't really transactions with Regtest, unless of course, if someone set up a fancy Regtest network for themselves.

I think this PR is good because:

1. Simple changes
2. Adds a feature for others
3. Doesn't impede on any other features.

cmd/bridgenode/chain.go
```diff
@@ -4,14 +4,16 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/mit-dci/lit/wire"
+
```
kcalvinalvin: Please remove the empty line here

ysangkok:
@kcalvinalvin all fixed!

kcalvinalvin:
@ysangkok I... still see the same code

ysangkok:
@kcalvinalvin sorry, for got to "git add -u". fixed now!

kcalvinalvin:
Could you also get rid of the empty line in ```cmd/bridgenode/chainio.go``` line number 7

ysangkok:
Seems you are against empty lines between groups of dependencies, so I removed them all:

```
diff --git a/cmd/bridgenode/chainio.go b/cmd/bridgenode/chainio.go
index 0f69e34..26daa93 100644
--- a/cmd/bridgenode/chainio.go
+++ b/cmd/bridgenode/chainio.go
@@ -4,7 +4,6 @@ import (
        "os"
 
        "github.com/mit-dci/lit/wire"
-
        "github.com/mit-dci/utreexo/cmd/util"
        "github.com/mit-dci/utreexo/utreexo"
 )
diff --git a/cmd/csn/client.go b/cmd/csn/client.go
index b7d60d4..1ed1d03 100644
--- a/cmd/csn/client.go
+++ b/cmd/csn/client.go
@@ -6,7 +6,6 @@ import (
        "time"
 
        "github.com/mit-dci/lit/wire"
-
        "github.com/mit-dci/utreexo/cmd/util"
        "github.com/syndtr/goleveldb/leveldb"
        "github.com/syndtr/goleveldb/leveldb/opt"
diff --git a/cmd/main.go b/cmd/main.go
index c65781d..e28d0ed 100644
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -8,7 +8,6 @@ import (
        "syscall"
 
        "github.com/mit-dci/lit/wire"
-
        bridge "github.com/mit-dci/utreexo/cmd/bridgenode"
        "github.com/mit-dci/utreexo/cmd/csn"
 )
```

kcalvinalvin:
> Seems you are against empty lines between groups of dependencies, so I removed them all:

We should try to adhere to the golang formatting and there should only be an empty line between the standard lib and external imports.

For future reference, if you run ```gofmt "filename"``` it'll output formatting code per golang standards.

ysangkok:
ah, right, thanks for the info! i will use a git commit hook. now `gofmt` doesn't report anything, i have indented the switches less.

ysangkok:
i have fixed a comment and added a new one:

```
diff --git a/cmd/util/utils.go b/cmd/util/utils.go
index 7e09cfc..2e7d933 100644
--- a/cmd/util/utils.go
+++ b/cmd/util/utils.go
@@ -151,7 +151,7 @@ func BtU32(b []byte) uint32 {
        return i
 }
 
-// uint32 to 4 bytes.  Always works.
+// int32 to 4 bytes (Big Endian).  Always works.
 func I32tB(i int32) []byte {
        var buf bytes.Buffer
        binary.Write(&buf, binary.BigEndian, i)
@@ -171,6 +171,7 @@ func BtI32(b []byte) int32 {
        return i
 }
 
+// uint32 to 4 bytes (Little Endian).  Always works.
 func U32tLB(i uint32) []byte {
        b := make([]byte, 4)
        binary.LittleEndian.PutUint32(b, i)
```

kcalvinalvin:
LGTM :)

kcalvinalvin:
@adiabat Please review this PR

PR #93, opened 2020-03-06
Fix spelling using codespell, run vet

ysangkok:
The vet changes a wasted Errorf call to a Printf,
so this change is not just cosmetic.


kcalvinalvin:
LGTM

@adiabat Please review and merge

PR #95, opened 2020-03-12
blockproof_test: Test proof stops being valid

ysangkok:
In a two leaf tree, we prove one node, then delete the other.

This commit adds a test asserts that the proof for the first node
stops being valid after deletion of the other node.

### DIFF
--- a/utreexo/blockproof_test.go
+++ b/utreexo/blockproof_test.go
@@ -1,6 +1,9 @@
 package utreexo
 
-import "testing"
+import (
+	"fmt"
+	"testing"
+)
 
 // TestVerifyBlockProof tests that the computedTop is compared to the top in the
 // Utreexo forest.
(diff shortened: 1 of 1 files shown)


kcalvinalvin:
I did go on a ramble on irc, making it harder to actually understand but the proofs should *fail* after deletion.

This is because now the proof for that particular leaf is different.

In a tree with 2 leaves, the tree looks like
```
02
|---\
00  01
```
In a tree like such, the proof for 01 would be:

1) Hash of 00
2) Hash of 01

We would then verify this proof by hashing those two items provided, and see if the result of the hash matches that of leaf 02.

**When 00 is deleted, the proof changes**

If the ```Modify()``` method is called and the user wants to delete 01, then the tree would transform and it would now be like such

```
01
```

It would be a tree with one leaf. So now, the proof is:

1) Hash of 01

And we would verify this by checking if the given proof matches the hash of 01 we have.


I mentioned adding the test since we should check that the *proof fails* after the deletion. This would be a sanity check to see if everything is going right.

Please do change this if statement ```if (!f.VerifyBlockProof(blockProof)) {``` so that it passes the test if the proof fails.

PR #96, opened 2020-03-13
forest_test: Add test deleting from empty tree

ysangkok:
Just a simple test asserting that we get an error
if we try to delete a leaf from an empty tree.

This test is currently consuming a lot of memory and does not pass.

### DIFF
--- a/utreexo/forest.go
+++ b/utreexo/forest.go
@@ -353,10 +353,10 @@ func (f *Forest) addv2(adds []LeafTXO) {
 // adds, which show up on the right.
 // Also, the deletes need there to be correct proof data, so you should first call Verify().
 func (f *Forest) Modify(adds []LeafTXO, dels []uint64) (*undoBlock, error) {
-	numdels, numadds := uint64(len(dels)), uint64(len(adds))
-	delta := numadds - numdels // watch 32/64 bit
+	numdels, numadds := len(dels), len(adds)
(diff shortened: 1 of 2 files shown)


ysangkok:
Causes underflow:

```
(dlv) 
> github.com/mit-dci/utreexo/utreexo.(*Forest).Modify() ./utreexo/forest.go:359 (PC: 0x5128a9)
   354:	// Also, the deletes need there to be correct proof data, so you should first call Verify().
   355:	func (f *Forest) Modify(adds []LeafTXO, dels []uint64) (*undoBlock, error) {
   356:		numdels, numadds := uint64(len(dels)), uint64(len(adds))
   357:		delta := numadds - numdels // watch 32/64 bit
   358:		// remap to expand the forest if needed
=> 359:		for f.numLeaves+delta > 1<<f.height {
   360:			// fmt.Printf("current cap %d need %d\n",
   361:			// 1<<f.height, f.numLeaves+delta)
   362:			err := f.reMap(f.height + 1)
   363:			if err != nil {
   364:				return nil, err
(dlv) p delta
18446744073709551586
(dlv) p numadds
0
(dlv) p numdels
30

```

kcalvinalvin:
Could you close #97 and merge that commit to this PR?

Looked at the test. Good catch. LGTM

ysangkok:
@kcalvinalvin  done

PR #97, opened 2020-03-13
forest: Avoid integer underflow

ysangkok:
When `len(dels)` > `len(adds)`, there would be an integer underflow,
which would make the loop condition always true, leading to a lot of
memory consumption when executing the test in #96.

This commit does not fix that test, but it does avoid the underflow
and the memory consumption.


kcalvinalvin:
Fix is good :)

As mentioned in #96, I think it'd be better organization wise if this commit was included in that PR.

PR #98, opened 2020-03-17
change positionMap into an interface

adiabat:
This sets it up so that forest can use levelDB for storing positions, and also have more than just position data associated with each hash (the utxo data needed for transaction verification)

Currently only is an interface for the same map functionality.
Seems to work, will test a bit more then merge if OK.


adiabat:
added default levelDB; everything seems to work; sync'd testnet OK.  Haven't tested but probably a little slower because of levelDB.  Right now defaulting to /dev/shm which is in ram so no big difference; can have position db path as an argument just like the forest data structure.
This should simplify, or really eliminate, the "restores the positionMap" part of RestoreForest()

Also if there's a levelDB we can get rid of miscForestFile and just use the position DB for that stuff.

The tests fail with some dependency thing but works on my end so I'm not sure what to change there.

ysangkok:
@adiabat yeah, the dependencies fetching broke when clair was merged. i'll prepare a PR for it before monday...

adiabat:
OK no problem.  Also I haven't fixed the undo tests so it'll still give a red x either way :)

ysangkok:
OK, I ran genproofs/ibdsim on testnet up to around block 175000, didn't error or anything.

adiabat:
Merging & will start using the `extra` field next

PR #99, opened 2020-03-18
Refactor cmd/ and add rev*.dat block reading capability to util/

kcalvinalvin:
Please do test out the rev_test.go file. You do need to copy rev*.dat files to the util/ directory for the test to work.

### DIFF
--- a/cmd/csn/server.go
+++ b/cmd/csn/server.go
@@ -5,7 +5,8 @@ import (
 	"os"
 	"time"
 
-	"github.com/mit-dci/lit/wire"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
 	"github.com/mit-dci/utreexo/cmd/ttl"
 	"github.com/mit-dci/utreexo/cmd/util"
 	"github.com/mit-dci/utreexo/utreexo"
@@ -16,7 +17,7 @@ import (
 // All the inputs are saved as 32byte sha256 hashes.
 // All the outputs are saved as LeafTXO type.
 func genPollard(
-	tx []*wire.MsgTx,
+	tx []*btcutil.Tx,
 	height int32,
 	totalTXOAdded *int,
 	lookahead int32,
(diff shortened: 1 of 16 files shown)


cmd/csn/server.go
```diff
@@ -27,28 +28,75 @@ func genPollard(
...
+	if err != nil {
+		return err
+	}
+	/*
+		if len(bp.Targets) > 0 {
+			fmt.Printf("block proof for block %d targets: %v\n", height+1, bp.Targets)
+		}
+	*/
+
+	// Fills in the niceces for verification/deletion
```
ysangkok: there is a typo here in nieces

cmd/csn/server.go
```diff
@@ -27,28 +28,75 @@ func genPollard(
...
+	// possible
```
ysangkok: what is possible?

cmd/util/compress.go
```diff
@@ -0,0 +1,588 @@
+package util
```
ysangkok: maybe it would make sense to write the commit_id of the btcd tree you took it from? so that people can easily know approximately how old it is

cmd/csn/server.go
kcalvinalvin: Whoops. Will fix

cmd/csn/server.go
kcalvinalvin: Will fix

cmd/util/compress.go
kcalvinalvin: good idea

ysangkok:
Hmm, isn't validate-4 supposed to be one commit ahead of validate-3? Somehow the latest commit hash in validate-3 isn't the second most recent commit in validate-4, even though they have the same diff and commit descriptions.

ysangkok:
I tested it on testnet up to around block 75000, didn't crash. But the revdata folder is empty after genproofs/ibdsim, is that expected?

kcalvinalvin:
> Hmm, isn't validate-4 supposed to be one commit ahead of validate-3? Somehow the latest commit hash in validate-3 isn't the second most recent commit in validate-4, even though they have the same diff and commit descriptions.

Ah yeah I just cherry-picked the commits from validate-3 to validate-4. Validate-4 is one commit ahead because of some other changes for txvalidation that isn't currently building yet.

> I tested it on testnet up to around block 75000, didn't crash. But the revdata folder is empty after genproofs/ibdsim, is that expected?

Ah yes. revdata/ directory is being created be cause of MakePaths() in util.go but the actual functions aren't yet called in genproofs. revdata/ should be empty.

PR #100, opened 2020-03-22
Fix CI

ysangkok:
- Also run for new pull requests
- More verbose during testing
- Don't use '...', clair breaks it
- Do not support 'dep', we will not use it

NOTE: Tests are still failing! So this will look red. It should be merged anyway.


adiabat:
LGTM!

PR #101, opened 2020-03-26
Fix undo tests

adiabat:
Seems like the main problem was the `cleanup` function... getting rid of it fixed all the problems with undo.

There are still a whole bunch of useless reads and writes of empty data (32 bytes of 0s) happening, so could look into those and stop that to speed up i/o.  But they don't look like they break anything.


kcalvinalvin:
noice

PR #107, opened 2020-04-01
cmd/: Refactor and simplify code

kcalvinalvin:
Main changes:

1. Simplified generating BlockProof and LeafTXO logic in both csn/ and bridgenode/
2. Asynchronous and buffered writes to pFile and pOffestFile
3. Simplified the dblookup logic and made it into a single function in
server.go. Deleted ttlparse.go and renamed ttldb.go to ttl.go. leveldb
ttl value lookup is still asynchronous with ibdsim but that logic is now
in genAdds() in server.go.
4. 10 second timeout added when given SIGINT, SIGQUIT, SIGTERM.


adiabat:
Looks good, I will start shuffling more things around from here

PR #108, opened 2020-04-02
WIP use rev____.dat files to recreate leaf data

adiabat:
Should work!  Currently doesn't; I think the rev reader isn't working properly, because it's making empty pubkeys


kcalvinalvin:
oof ok I'll look at the rev reader

adiabat:
OK this seems to work.  It's not quite finished as it needs some accumulator changes, but does work without them (just inefficiently)

I get one red x and one green check, not sure what that means.  The red x is a test that fails at forest_test.go:293:24, but forest_test.go only has 161 lines so it's probably fine.

PR #110, opened 2020-04-06
Add test deleting keys 1,0 from two-leaf tree

ysangkok:
This is causing a panic on master.

I have let the commit message note the commit ID of master as it is currently, so that the fix can be merged first, if you like. That way, the commit description will still make sense.

Note that if you add another leaf, it still panics with the same error message. I am saying this because I suspected it was because of emptying the tree. But it seems it wasn't.

### DIFF
--- a/utreexo/forest_test.go
+++ b/utreexo/forest_test.go
@@ -5,6 +5,21 @@ import (
 	"testing"
 )
 
+func TestDeleteReverseOrder(t *testing.T) {
+	f := NewForest(nil)
+	leaf1 := LeafTXO { Hash: Hash { 1 } }
+	leaf2 := LeafTXO { Hash: Hash { 2 } }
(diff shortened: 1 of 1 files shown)


adiabat:
OK - I had put this off because in practice there will never be 0 utxos, but yeah this should work & be supported.  Will try to have my next PR make this test pass!

ysangkok:
But the error occurs even if you have one more leaf, and still delete only these two, that were previously there (as I wrote in the original message)!

adiabat:
Looking, the problem is `f.Modify(nil, []uint64{1, 0})`
Modify requires that the list of deletions is sorted, and if it's unsorted it panics on "got non-moving swap".  That should probably be an error instead of a panic, or maybe Modify should just check that it's sorted or sort it in the beginning.

PR #111, opened 2020-04-09
Add a test for random small forests

ysangkok:
These forests have up to 256 leaves, generated using testing/quick.

The test is currently triggering another panic.


adiabat:
Yup - I think the problem is in forest.positionMap.
I kind of want to fix it before merging the test, then everything will have green checkmarks :)

PR #112, opened 2020-04-18
Start converting to log module instead of Println

ysangkok:
In this PR are only logging categories for CSN and Pollard, but more can be added if you think the approach is OK.

### DIFF
--- a/utreexo/pollardutil.go
+++ b/utreexo/pollardutil.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/mit-dci/utreexo/cmd/util"
+	"github.com/mit-dci/utreexo/log"
 )
 
 // Pollard is the sparse representation of the utreexo forest, using
@@ -16,6 +17,8 @@ import (
 
 // Pollard :
 type Pollard struct {
+	loggers log.Loggers
+
 	numLeaves uint64 // number of leaves in the pollard forest
 
 	tops []polNode // slice of the tree tops, which are polNodes.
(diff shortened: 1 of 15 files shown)


kcalvinalvin:
Getting a build error with this log on my local machine and github actions confirms this.

```
../utreexo/pollardutil.go:8:2: cannot find package "github.com/mit-dci/utreexo/log" in any of:
	/usr/local/Cellar/go/1.13.6/libexec/src/github.com/mit-dci/utreexo/log (from $GOROOT)
	/Users/timapple/go/src/github.com/mit-dci/utreexo/log (from $GOPATH)
```

ysangkok:
@kcalvinalvin  Yeah, I dunno if I am doing something wrong, but the log module needs to be available to utreexo in the future too, so that's why it is on the level with cmd and utreexo. You can build by running `go build` in side the `cmd` (as your working directory)

kcalvinalvin:
> You can build by running `go build` in side the `cmd` (as your working directory)

I should note that I did this when testing and go gave the build error.

utreexo/pollardutil.go
```diff
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/mit-dci/utreexo/cmd/util"
+	"github.com/mit-dci/utreexo/log"
```
kcalvinalvin, line 8: Did you add a package to the repo? In the commit, there is no log/ directory

ysangkok: Oh, right, that explains the failure too. Sorry, dunno how I could have missed this.

Issue #114, opened 2020-04-21
Runtime error in current master


kcalvinalvin:
```
leveldb: closed
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/mit-dci/utreexo/cmd/bridgenode.genDels(0x17c, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x9048b9c500000001, ...)
        /home/calvin/go/src/github.com/mit-dci/utreexo/cmd/bridgenode/genproofs.go:300 +0x42b
github.com/mit-dci/utreexo/cmd/bridgenode.genAddDel(0x17c, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x9048b9c500000001, ...)
        /home/calvin/go/src/github.com/mit-dci/utreexo/cmd/bridgenode/genproofs.go:278 +0x8e
github.com/mit-dci/utreexo/cmd/bridgenode.BuildProofs(0xc00709110b, 0x5f8f4a, 0xd, 0x5fa4e5, 0x12, 0xc00045c000, 0x0, 0x0)
        /home/calvin/go/src/github.com/mit-dci/utreexo/cmd/bridgenode/genproofs.go:111 +0x84a
main.main()
        /home/calvin/go/src/github.com/mit-dci/utreexo/cmd/main.go:69 +0x2be
```

I'm currently getting this index range error when tested the ```cmd``` binary on testnet

adiabat:
Huh, not sure what's going on.  I tried it with testnet and mainnet blk/rev files and I can't get this same bug to happen.

Can you try to find what it's crashing on?  Is this right at the beginning, or a bunch of blocks in?  Also some info about what transaction it's crashing on, maybe it's something weird.  

Since the tests all work but it crashes on actual use, it'd be great if you could write a test that causes the same crash.

My guess it's some off-by-1 thing.  In `GenDels()` `txinblock` is decremented and should be the same as the index of the rev transactions, but might not be in some case.  I could add a check there and return an error if there isn't 1 fewer rev tx than regular tx.  But maybe this happens in rev blocks sometimes and it's not an error, so we have to deal with it some other way.

ysangkok:
My logs for testnet:

```
Building offsetfile... blk00181.dat
got non magic bytes 00000000, finishing
Building offsetfile... blk00182.dat
blk00182.dat doesn't exist; done building
tip height 1720819
[...]

PR #116, opened 2020-04-24
util/utils.go: Panic when rev*.dat files are not present

kcalvinalvin:
This will let users know that they're missing rev*.dat files


adiabat:
Hm this might also be one of the ignored problems related to #114 ..?

PR #117, opened 2020-04-27
Move folders and packages around

adiabat:
I think this makes more sense; the "utreexo" package is now called "accumulator" which is what it contains.  "cmd" now only has the main package, and csn / bridge are in their own folders.

Once networking is in, I think it makes sense to have 2 different binaries, which is somewhat client / server-ish but probably easier to deal with now.


ysangkok:
Looks sensible! What are the doubts?

adiabat:
OK merging this; I got it to stop giving red xs, but also no checkmarks, so I guess I broke the CI thing.  Anyway works on my computer.

PR #118, opened 2020-05-01
Fix YAML syntax in Github test configuration

ysangkok:
This makes the tests run again. Tested in my fork.


adiabat:
Aha!  Put them all on the same line, I get it...

PR #121, opened 2020-05-07
WIP client / server networking

adiabat:
the genproofs side listens, and the ibdsim side connects.
Should be pretty straightforward as both sides are already using worker routines to get data off disk and put it in a channel; now it's just the function that feeds the channel is getting the data over the network instead of from the disk.  
Initially it will be real simple, just getting everything in order, with no real messages about which blocks to get.


adiabat:
Will merge for now, and continue to fix up network functions.  (like stop / resume.  Then maybe caching)

PR #122, opened 2020-05-09
DRY in transform (new function swapInRow)

ysangkok:
Also removing the branching on mask since branching is more expensive and complicated than just doing the zero-XOR.


adiabat:
Looks good, thanks!  Makes sense that XOR is cheaper than branch.

PR #125, opened 2020-05-17
transform: Split up remTrans2

ysangkok:
Might even be a bit faster since arrays are initialized to their correct size.

### DIFF
--- a/accumulator/transform.go
+++ b/accumulator/transform.go
@@ -23,7 +23,7 @@ func remTrans2(dels []uint64, numLeaves uint64, forestRows uint8) [][]arrow {
 			break
 		}
 
-		var twinNextDels, swapNextDels []uint64
+		var twinNextDels []uint64
 
 		// *** Delroot
 		// TODO would be more elegant not to have this here.  But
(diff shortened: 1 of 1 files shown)


kcalvinalvin:
So the code is behaving the same as current master (testnet3) but I think it's missing some edge cases. I'm currently only able to sync up to block 85,000 as well so I can't really do an exhaustive test at the moment.

It's also combining the steps ```swap``` and ```root``` which really shouldn't be since ```root step```'s role is to clean up the remaining single leaf after the ```swap step```.

In the original code, ```collapses``` can be assigned in either 

```
if rootPresent && !delRemains 
```

or 

```
if !rootPresent && delRemains
```

But in the PR, there's no checks for if there are delremains
```
	if rootPresent {
		rootSrc = rootPosition(numLeaves, r, forestRows)
	} else {
		// no root but 1 del: sibling becomes root & collapses
		// in this case, mark as deleted
		rootSrc = dels[len(dels)-1] ^ 1
```

Overall the 3 new function that are defined are splitting up the transform ```root step```.  The ```root step``` is defined as such. This could really be a switch/if else statements as only one of these cases are gonna run.
https://github.com/mit-dci/utreexo/blob/5075374f3f311977b2e262e81739da49278c3ac6/accumulator/transform.go#L58-L79

Also, this code segment in ```makeSwapNextDels()``` and ```makeSwaps()```wouldn't really be necessary since root is for processing a single leaf that didn't have a pair in the ```swap``` step of transform.
```
	delRemains := len(dels)%2 != 0
	numSwaps := len(dels) >> 1
	if delRemains && !rootPresent {
		numSwaps++
	}
	swapNextDels := make([]uint64, numSwaps)
```

The code does run but I'm not sure if it behaves the same in these edge cases for the ```root step```.

@adiabat please review this PR

accumulator/transform.go
```diff
@@ -1,5 +1,65 @@
 package accumulator
 
+func makeCollapse(dels []uint64, rootPresent bool, r uint8, numLeaves, nextNumLeaves uint64, forestRows uint8) []arrow {
+	delRemains := len(dels)%2 != 0
+	if rootPresent == delRemains {
```
ysangkok, line 5: @kcalvinalvin Did you see this line? This makes sure the function only proceeds in the desired configurations of rootPresent and delRemains, keeping the old behaviour. I chose to do it this way, because the loop is already complicated, so I preferred to have branching in the functions, which are pure.

ysangkok:
> Also, this code segment in makeSwapNextDels() and makeSwaps() wouldn't really be necessary

I don't understand your suggestion? This PR is merely shuffling code into pure functions, the numSwaps++ line you suspect not to be necessary is implementing the behaviour of line 59-61 in your quoted master segment.

ysangkok:
@kcalvinalvin I have added a commit using the switch statement you suggested. Since you missed the early return, it must have been hard to read. This makes the three branches more obvious.

accumulator/transform.go
kcalvinalvin: oh I see what's going on now. Yeah I did miss the line :/

kcalvinalvin:
> @kcalvinalvin I have added a commit using the switch statement you suggested. Since you missed the early return, it must have been hard to read. This makes the three branches more obvious.

Thanks for the change. Lot harder to misunderstand it now :)



> > Also, this code segment in makeSwapNextDels() and makeSwaps() wouldn't really be necessary
> 
> I don't understand your suggestion? This PR is merely shuffling code into pure functions, the numSwaps++ line you suspect not to be necessary is implementing the behaviour of line 59-61 in your quoted master segment. the last line of
> 
> https://github.com/mit-dci/utreexo/blob/5075374f3f311977b2e262e81739da49278c3ac6/accumulator/transform.go#L73-L78

Ah ok I also see what's happening here now. Didn't see that ```dels``` from extractTwins was being fed to both ```makeSwaps()``` and ```makeSwapNextDels()```.

I guess one request would be to comment the new functions and the code inside so it's easier to read. There are old comments that were carried over that should be deleted too.

Rest of the code LGTM

adiabat:
Thanks, I think this can be easier to read than having one big loop.  Things I think would improve readability:

- I guess keep `remTrans2()` at the top and the smaller functions beneath.  Not like it matters to the computer at all.

- I'd keep `delRemains` calculated and defined before doing anything else in the loop, and then passing the bool to the functions.  That way it's clearer that delRemains is known right at the start of the row.

- The default: panic in the switch statement seems strange, because just looking at 3 lines (12-15) it's clear it can never get there.  So I'd be fine leaving it empty.  Or making default do nothing, and having it be the `rootPreset == delRemains` case.  That's maybe easier to understand, as it makes the other case statements more explicit ( `case rootPresent && !delRemains`, `case !rootPresent && delRemains`)

- Maybe there's some way to not calculate `rootPos` if `rootPresent == false`?  It's weird to have the invalid `rootPos` variable floating around.  It would also let you throw a panic in the `rootPosition()` util function.  I don't see a nice way to do this though, it looks like it could instead make it harder to read, so maybe not.

- I'd keep the `// *** Twin` comments over the function calls... or really I can re-arrange the comments, a lot of them don't fit.

The only major thing is to keep `delRemains` and pass it, as I think that's clearer.  Otherwise, yeah I'm pretty convinced it does the same thing & I'm OK to merge it.

adiabat:
OK looks good, hope this makes it easier for people to understand.

PR #126, opened 2020-05-22
Forest: Split up removev4

ysangkok:
A bunch of commented code and hDestSlice was also removed. hDestSlice was equal to nextHashDirt as far as I could observe.


adiabat:
Tested with testnet3 data and works.  Looks good, this seems easier to understand, thanks!

PR #127, opened 2020-05-22
bridgenode: accept connections in separate goroutine

dergoegge:
I saw this comment:
https://github.com/mit-dci/utreexo/blob/23d9af597ccf5175527483667e112d04fcdb7c67/bridgenode/server.go#L38-L40

and thought: why not move the accepting of new connections to its own goroutine and pass connections back to the select statement via a channel? that way the arbitrary timeout is not needed.

disclaimer: i did not test this, since i currently don't have a full node setup.

### DIFF
--- a/bridgenode/server.go
+++ b/bridgenode/server.go
@@ -4,7 +4,6 @@ import (
 	"encoding/binary"
 	"fmt"
 	"net"
-	"time"
 
 	"github.com/mit-dci/utreexo/util"
 )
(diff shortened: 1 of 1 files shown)


adiabat:
This is a much better way to do it, thanks!

I was looking online and all I found was the timeout idea, which struck me as not the best way to do it -- pushing the connections themselves into a channel seems much better & more go-like.

I tested it out with testnet and it works. :+1:

Issue #103, opened 2020-03-30
Data race in utreexo/hasher.go


kcalvinalvin:
Building with go build -race and running ibdsim gives me a data race warning. Not sure how serious this is.

```
==================
WARNING: DATA RACE
Write at 0x00c000304c60 by goroutine 38:
  github.com/mit-dci/utreexo/utreexo.(*hashableNode).run()
      /Users/timapple/go/src/github.com/mit-dci/utreexo/utreexo/hasher.go:20 +0x1ad

Previous write at 0x00c000304c60 by goroutine 24:
  github.com/mit-dci/utreexo/utreexo.(*hashableNode).run()
      /Users/timapple/go/src/github.com/mit-dci/utreexo/utreexo/hasher.go:20 +0x1ad

Goroutine 38 (running) created at:
  github.com/mit-dci/utreexo/utreexo.(*Pollard).rem2()
      /Users/timapple/go/src/github.com/mit-dci/utreexo/utreexo/pollard.go:226 +0xe8c
  github.com/mit-dci/utreexo/utreexo.(*Pollard).Modify()
      /Users/timapple/go/src/github.com/mit-dci/utreexo/utreexo/pollard.go:10 +0x5d
  github.com/mit-dci/utreexo/cmd/csn.genPollard()
      /Users/timapple/go/src/github.com/mit-dci/utreexo/cmd/csn/server.go:59 +0x2d4
  github.com/mit-dci/utreexo/cmd/csn.IBDClient()
      /Users/timapple/go/src/github.com/mit-dci/utreexo/cmd/csn/client.go:80 +0x4f3
  github.com/syndtr/goleveldb/leveldb/storage.(*fileStorage).List()
      /Users/timapple/go/src/github.com/syndtr/goleveldb/leveldb/storage/file_storage.go:458 +0x2c4
  fmt.Fscanf()
      /usr/local/Cellar/go/1.13.6/libexec/src/fmt/scan.go:143 +0xcf
  github.com/syndtr/goleveldb/leveldb/storage.fsParseName()
      /usr/local/Cellar/go/1.13.6/libexec/src/fmt/scan.go:114 +0x192
  github.com/syndtr/goleveldb/leveldb/storage.(*fileStorage).List()
      /Users/timapple/go/src/github.com/syndtr/goleveldb/leveldb/storage/file_storage.go:458 +0x2c4
  fmt.Fscanf()
[...]

PR #129, opened 2020-05-25
accumulator: make dedupeSwapDirt work as documented

dergoegge:
The util function `dedupeSwapDirt` currently does not work as intended/documented.
https://github.com/mit-dci/utreexo/blob/cb5c021b3d001a5f5b4dd3eedce8aac034a5dbd2/accumulator/utils.go#L389-L392

Example:
```go
dedupeSwapDirt([]uint64{1, 4, 5, 10, 16}, []arrow{{1, 4}, {1, 11}, {1, 16}})
```
This currently results in `[1 4 5 10 16]` but it should result in `[1, 5, 10]`.

### DIFF
--- a/accumulator/utils.go
+++ b/accumulator/utils.go
@@ -400,7 +400,7 @@ func dedupeSwapDirt(a []uint64, b []arrow) []uint64 {
 	idxb := 0
 	for j := 0; j < maxa; j++ {
 		// skip over swap destinations less than current dirt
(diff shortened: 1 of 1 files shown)


kcalvinalvin:
@adiabat perhaps this fixes #103. Will test soon

kcalvinalvin:
Doesn't fix #103 but it works as described.

PR #131, opened 2020-05-25
Remove complicated branching in makeDestInRow

ysangkok:
I must admit I don't understand the logic for this complicated branching logic. But I thought I'd propose this simplification. If it can't be simplified, would be nice to explain the logic.


adiabat:
This part isn't commented well and *seems* to do something in that this change does make it sometimes return true instead of false.  But from testing it seems the same hashes are computed, so it's just a performance change.  Probably related to hashing things when we don't need to, which is also an issue.  
Simplified unless we know why the complexity is there sounds OK to me :)

PR #132, opened 2020-05-27
Add csn/bridgenode integration test

dergoegge:


test/csn_bridge.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
+# Integration test script for the csn and bridge node.
+# WARNING: The bitcoin config at $BITCOIN_CONF will be overridden.
```
ysangkok, line 3: could use mktemp to avoid this?

dergoegge: oh yes cool, thanks

test/csn_bridge.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/bash
+# Integration test script for the csn and bridge node.
+# Assumes that `the following binaries exist:
+# 	`bitcoind`, `bitcoin-cli, `utreexo`
+
```
ysangkok, line 5: would be cleaner to use `set -euo pipefail`. see more at https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

adiabat:
I had trouble getting this to work on my computer.  I had to change the binary calls to absolute paths.  For example in `run_utreexo()` :
```utreexo genproofs -net=regtest > $TEST_DATA/genproofs.log 2>&1 &```
didn't work because I'd need a utreexo binary in PATH installed
```./cmd genproofs -net=regtest > $TEST_DATA/genproofs.log 2>&1 &```
doesn't work either for reasons I'm not sure, even though I had the compiled cmd binary in the same directory, it said `./csn_bridge.sh: line 91: ./cmd: No such file or directory`
Using absolute paths like this:
```/home/adiabat/utreexotest/cmd genproofs -net=regtest > $TEST_DATA/genproofs.log 2>&1 &```
works though.  Maybe because it's going into a temporary directory so relative paths don't work.

Once I changed those, it would run, but it hung on `waiting for genproofs to start the blocks server...`
that may be a problem with genproofs, or a directory based problem.

One way I'd suggest to help with this is make the binary names and paths variables near the top of the script so if it's called `cmd` or `utreexo` it's easy to change in one place, and put a path there.  I think in many cases the utreexo binaries won't be in the shell path so we need to put a full path to that.  `bitcoind` and `bitcoin-cli` are often in the shell's path but we can also have those called with absolute paths to be safe.  The paths to the programs could also be arguments to `csn_bridge`

Getting stuck on waiting for genproofs might be something path related but I'm not sure, have you seen that?

**Comment by dergoegge** [comment]
The reason why the relative path doesn't work is because the script cds to the bitcoin block dir before executing utreexo.
https://github.com/mit-dci/utreexo/blob/6c1c2f6f3a2086416216954e8af611cea6d18a48/test/csn_bridge.sh#L142

Passing the path to the programs as arguments is a good idea, that way it won't matter when the cmd binary gets split into `csn` and `bridge`.

My best guess as to why you get stuck on `waiting for genproofs to start the blocks server...` is that you don't have netcat as `nc` on your system. netcat is used to check if port 8338 is open on localhost.
https://github.com/mit-dci/utreexo/blob/6c1c2f6f3a2086416216954e8af611cea6d18a48/test/csn_bridge.sh#L97-L99

I will add some checks at the begining to check if all the binaries actually exist.

File: test/csn_bridge.sh [hunk 3]
```diff
@@ -0,0 +1,186 @@
+#!/bin/bash
```
kcalvinalvin, line 1: We should prefer ```#!/usr/bin/env bash``` since absolute paths can fail.

kcalvinalvin:
Could probably rip this bitcoind install script https://github.com/lightningnetwork/lnd/blob/master/scripts/install_bitcoind.sh

PR #133, opened 2020-05-28
Fix regtest panic

dergoegge:
* Regtest has its own magic bytes
* only 28 bytes were read when reading previous blockhashes from blk.dat files.


adiabat:
Cool thanks, was not aware of different regtest magic bytes

PR #134, opened 2020-05-28
consensus:Add initial support for tx script validation

kcalvinalvin:
This works except for the fact that UtxoViewPoint.entry is unexported.


kcalvinalvin:
Closing in favor of 135

PR #135, opened 2020-05-28
consensus: Utxoview based script validation

adiabat:
It's a different way of doing #134 and doesn't do all the checks #134 has, but seems to work with minimal additions to btcd.


kcalvinalvin:
```
Creating new pollarddata                                                                                 │
panic: EOF                                                                                               │
                                                                                                         │
goroutine 8 [running]:                                                                                   │
github.com/mit-dci/utreexo/util.UblockNetworkReader(0xc00006c1e0, 0x65167f, 0xe, 0x3e800002103)          │
        /home/calvin/bitcoin-projects/utreexo/go/src/github.com/mit-dci/utreexo/util/utils.go:159 +0x2ac │
created by github.com/mit-dci/utreexo/csn.IBDClient                                                      │
        /home/calvin/bitcoin-projects/utreexo/go/src/github.com/mit-dci/utreexo/csn/client.go:52 +0x19d  │
Command exited with non-zero status 2                                                                    │
57.99user 9.07system 0:23.06elapsed 290%CPU (0avgtext+0avgdata 42616maxresident)k                        │
0inputs+0outputs (0major+11973minor)pagefaults 0swaps    
```

I'm getting this from ibdsim side and ```pushBlocks GetUDataFromFile binary.Read offset 8450 EOF``` from the genproofs side. Not quite sure if this is the expected behavior.

util/udata.go
```diff
@@ -2,6 +2,16 @@ package util
 
 import (
 	"fmt"
+
+	"github.com/btcsuite/btcd/txscript"
```
kcalvinalvin, line 6: The extra empty lines in imports shouldn't be there

dergoegge:
> I'm getting this from ibdsim side and `pushBlocks GetUDataFromFile binary.Read offset 8450 EOF` from the genproofs side. Not quite sure if this is the expected behavior.

@kcalvinalvin This should be the same thing i ran into, that happens when genproofs is done sending blocks and closes the connnection. Got fixed with #136.

adiabat:
I'm having some build weirdness on my local machine but I probably just don't have go mod set up right.  Seems to work OK here though so merging.  It still doesn't do everything, but it does the important signature checks.
Also -- big TODO -- it doesn't 1 transaction at a time, so really need to change it to do all the transactions in parallel.

PR #136, opened 2020-05-28
csn: exit ibdsim when EOF is received

dergoegge:


adiabat:
Eventually we might want to keep in sync after getting to the tip, but exiting is certainly better than crashing :)

PR #137, opened 2020-05-29
consensus: Add initial support for tx script validation without utxoview

kcalvinalvin:
More code but no utxoview needed and no changes to btcd needed


kcalvinalvin:
closing in favor of #135

PR #139, opened 2020-05-29
docs: Update readme to mention that ibdsim requests blocks from

kcalvinalvin:


readme.md
```diff
@@ -78,4 +69,6 @@ $ ./cmd ibdsim -net=testnet # -net=testnet flag needed for testnet. Leave out fo
 [To resume, just do `./cmd ibdsim -net=testnet` again]
 ```
 
+* `ibdsim` is the CSN node and it will call genproofs and ask for blocks with the Utreexo accumulator proofs. It will receive the proofs and validate the inclusion.
```
ysangkok, line 72: since this is the first use of this abbreviation in this file, it should be expanded

kcalvinalvin: Maybe it's a good idea to explain ibd somewhere (or maybe it could be assumed that people already know?) but here I'm referring to the ibdsim argument with the binary.

ysangkok: yeah that too

adiabat:
OK looks good; I am splitting the cmd binary up into bridge and csn in the hook branch so I'll update this readme to reflect that once it gets split up (which I think will be cleaner but if you think it's better to keep one cmd let me know)

PR #141, opened 2020-05-30
util: Refactor integer conversion functions

dergoegge:
In an effort to reduce the util code...
The std library has all the functionality needed to convert integers to bytes and vice versa.

* In some cases this gets rid of the extra allocation, when converting to bytes.
* makes the code more readable (at least in my opinion)


kcalvinalvin:
Looks good :) Tested and reviewed

PR #143, opened 2020-06-03
updateDirt maintains sortedness

ysangkok:
Before this commit, updateDirt would return an unsorted slice,
which would then be passed back into updateDirt. Then,
dedupeSwapDirt would be called with its invariant violated.
So instead of the half-hearted checking of the last element,
we simply do a binary search (cheap) and insert the element
where it should be. No special treating of the last element.

### DIFF
--- a/accumulator/forest.go
+++ b/accumulator/forest.go
@@ -4,6 +4,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"os"
+	"sort"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


adiabat:
OK I think this is better than what's there now. 
I think it could be even better though, because sort.Search() doesn't know anything about the target it's looking for.
In this case the slice now stays sorted (cool!) the index we're looking for is going to be almost at the end.
It's log(n) reads but it'd be cool to somehow give it the hint to start at the end instead of the middle.
Maybe it's not a big deal because this gets called about as often as a hash needs to be computed, and the hash is probably a lot more expensive.  Anyway something to think about...

Issue #130, opened 2020-05-25
Panic while generating proofs for testnet


dergoegge:
```
On block : 190001
On block : 200001
remap forest 19 rows -> 20 rows
remap forest 20 rows -> 21 rows
panic: genDels block 206482 87 txs but 0 rev txs

goroutine 1 [running]:
main.main()
        /Users/niklas/go/src/github.com/mit-dci/utreexo/cmd/main.go:71 +0x525
```
I am getting this error/panic message on current master(https://github.com/mit-dci/utreexo/commit/cb5c021b3d001a5f5b4dd3eedce8aac034a5dbd2)  when running the `cmd` binary on testnet.

Might be related to #114 

Might also be caused by:
* *running on macOS*
* *syncing to testnet using the bitcoin core master branch(https://github.com/bitcoin/bitcoin/commit/dc5333d31f280e09bb1e8cdacfbe842f4ab9e69b), which is not guaranteed to be completely stable.*

kcalvinalvin:
Ah this is because of block order misalignment in ```blk*.dat``` files and ```rev*.dat``` files. Currently the master branch of Utreexo uses the ```rev*.dat``` files for fetching tx validation data for blocks but there's isn't *anything* in the ```rev*.dat```` file to indicate the height of the rev block.

So by the error message you've reproduced, it seems like the ```rev*.dat``` file is off order at height 206482. This is normal behavior that the current Utreexo code isn't handling right now.

We're aware and talked about the issue on the weekly meetings and going forward, instead of relying on ```rev*.dat``` files, we'll replace them with rpc calls to bitcoin core.

adiabat:
Yeah this is most likely the same error as #114 ... it's a tricky one to fix.
The rev.dat files are mostly in the same order as the blk.dat files, but in some cases it's not.  The tricky part is that just given the rev and blk files, there's no sure way to figure out the ordering of the rev data.  The blk files are at least a blockchain, where hashes point to the previous block, so even when they're out of order you can figure out the correct order from the hashes.  The rev blocks don't have hashes that mean anything though.

One way to fix this that will work like 99+% of the time would be to check that the block and rev block at the assumed same height match in terms of transactions and inputs, and if they don't keep the rev block in ram in a map until we can assign it.

[...]

PR #144, opened 2020-06-06
bridgenode: Read rev block offsets from index/ directory

kcalvinalvin:
Things done:
Commit 1
1. Now rev block offset is read from ```blocks/index/```.  Fixing #130
2. The rev offset is now appended to ```offsetdata/offsetfile```, getting rid of ```revoffsetdata/```.
3. Option ```datadir``` added to main so that the user can give a custom datadir directory.
4. Testing for building the offsetfile added in ```bridgenode/offsetfile_test.go```. Needs to be run manually for now since there needs to be the ```block/``` data

Commit 2
5. Updated workflows so that the fetching of dependencies is done from the code in the PR


kcalvinalvin:
Rev stuff works now :) Just need to put an option for datadir in the binary

bridgenode/rev.go
```diff
@@ -0,0 +1,348 @@
...
+// OpenIndexFile returns the db with only read only option enabled
+func OpenIndexFile() *leveldb.DB {
+	s := "/home/calvin/.bitcoin/testnet3/blocks/index"
+	lvdb, err := leveldb.OpenFile(s, nil)
+	if err != nil {
+		panic(err)
+	}
+
+	ro := new(opt.ReadOptions)
+	ro.DontFillCache = true
```
dergoegge, line 256: Is the db really opened with readonly? https://github.com/syndtr/goleveldb/blob/758128399b1df3a87e92df6c26c1d2063da8fabe/leveldb/opt/options.go#L609-L614

also seems like `ro` isn't used

kcalvinalvin: Ah yeah something to clean up before un-drafting. Thanks!

kcalvinalvin: addressed

dergoegge:
Tested with the block data used in #130 and it works :)

dergoegge:
I am getting this when i pause genproofs:
```
On block : 290001
^CUser exit signal received. Exiting...
numLeaves= 2605178
f.rows= 22
Done writing
Program timed out. Force quitting. Data likely corrupted
```

it can resume fine though, so the data is not actually corrupted.

kcalvinalvin:
> I am getting this when i pause genproofs:
> 
> ```
> On block : 290001
> ^CUser exit signal received. Exiting...
> numLeaves= 2605178
> f.rows= 22
> Done writing
> Program timed out. Force quitting. Data likely corrupted
> ```
> 
> it can resume fine though, so the data is not actually corrupted.

Ah yeah that's because ```blockServer()``` isn't receiving the sigint given by the user. This behavior is seen on master as well. Something to fix but not caused by this PR. https://github.com/mit-dci/utreexo/blob/cb337c59df3b8e0e7e6359b70965320b9c000d05/bridgenode/genproofs.go#L151

The problem is caused by channel ```haltRequest``` being consumed by the for loop at https://github.com/mit-dci/utreexo/blob/cb337c59df3b8e0e7e6359b70965320b9c000d05/bridgenode/genproofs.go#L130.

adiabat:
I'm getting an error
```Building offsetfile... blk00183.dat
got non magic bytes 00000000, finishing
panic: leveldb: not found

goroutine 1 [running]:
bridgenode.GetBlockIndexInfo
...utreexo/bridgenode/rev.go:300
```
So I guess it's trying to read a lookup height that's not there in the db.  This is with a just-sync'ed and stopped testnet3 from bitcoind 0.20.  Any ideas how to make it work?

kcalvinalvin:
> I'm getting an error
> 
> ```Building offsetfile... blk00183.dat
> 
> got non magic bytes 00000000, finishing
> 
> panic: leveldb: not found
> 
> 
> 
> goroutine 1 [running]:
> 
> bridgenode.GetBlockIndexInfo
> 
> ...utreexo/bridgenode/rev.go:300
> 
> ```
> 
> So I guess it's trying to read a lookup height that's not there in the db.  This is with a just-sync'ed and stopped testnet3 from bitcoind 0.20.  Any ideas how to make it work?

Could you check that you’re on the latest commit? It shouldn’t ever fail after building the offsetfile since the first thing it does when building the offsetfile is fetch data from leveldb.

Some of my old WIP commits had the offsetfile build and then fetch from db so I think this is it.

adiabat:
I'm on 206fe55d8a1bf529839db633de0434e141db01bd ... I thought it might be a datadir problem so I tried putting the files into `~/.bitcoin/testnet3/blocks` and running without -datadir but got the same thing

adiabat:
It looks like it's getting most of the way through the first rev block, and not finding `0000009aec10dbbf37982fc2d85e0d4b0e45ee3438d039f7c1cc6fe2f3041562` which is... not a block on testnet? at least according to block explorers.  Is the wrong hash getting in there somehow..?

adiabat:
OK revised - I think it's getting to the end of everything, not just the first file, but there may be blocks in blk.dat that aren't even in rev.dat.  Who knows.  I'd say instead of panicking just say you're done if this happens.  Because it's happening to me.

I re-ran bitcoind and got a new set of files that it didn't have this problem with.  Soooo... I guess I'll merge?  Because I over-wrote the old index and I bet I won't be able to reproduce that again without syncing testnet all over again...

kcalvinalvin:
> OK revised - I think it's getting to the end of everything, not just the first file, but there may be blocks in blk.dat that aren't even in rev.dat.  Who knows.  I'd say instead of panicking just say you're done if this happens.  Because it's happening to me.
> 
> 
> 
> I re-ran bitcoind and got a new set of files that it didn't have this problem with.  
> 
> Since the whole point of this PR is to fix genproofs to work reliably with the blk/rev.dat files, I think GetBlockIndexInfo shouldn't be able to panic if something's not in the db, because that happens.
> 
> 
> 
> Instead it can return a not found error, which writeBlockOffset can handle by finishing up at the not found block height.  In my case it was only the very last block that didn't have rev data, so probably that's of minimal impact when people want to try this out.

Huh. That is.... odd (maybe you got an orphaned block or something?). Ok will do so.

kcalvinalvin:
https://bitcoin.stackexchange.com/questions/92128/is-the-ordering-of-blocks-in-the-rev-dat-files-same-as-the-blk-dat-files

From this post, the undo blocks are written when a block is fully verified. If there is no undo block, it means it's not in the best chain. I think it's better to just ignore that block and print a warning that the user may be wasting disk space.

PR #145, opened 2020-06-09
bridgenode: Dont start the blockServer if genproofs was paused

dergoegge:
Fixes the problem mentioned here: https://github.com/mit-dci/utreexo/pull/144#issuecomment-640761200


adiabat:
Ah good catch, I noticed it wasn't breaking correctly and I guess I introduced that! oops :)

PR #147, opened 2020-06-12
accumulator: Add a cache for disk forest data

dergoegge:
Adds a new type `cacheForestData` that implements the `ForestData` interface. This type improves the performance of `diskForestData` by caching parts of the forest in ram.
The cache holds a certain number of trees of the full forest while the other trees are kept on disk, the trees kept in the cache are the smallest trees in the forest. The whole forest might be held in ram when the cache size allows it.

Usage:
```go
d := new(cacheForestData)
// 16 means that the 16 smallest trees will be cached.
// that's only about ~4MB, 17 would be ~8MB, 18 -> ~16MB, ..., 27 -> 8448MB, ...
d.cache = newDiskForestCache(16)
d.f = forestFile
```

### DIFF
--- a/accumulator/forest.go
+++ b/accumulator/forest.go
@@ -111,10 +111,7 @@ func NewForest(forestFile *os.File) *Forest {
 		d := new(diskForestData)
 		d.f = forestFile
 		f.data = d
-		d.cache = diskForestCache{
-			Size: 1 << 11,
-			data: make(map[uint64]Hash),
-		}
+		d.cache = newDiskForestCache(24)
 	}
 
 	f.data.resize(1)
@@ -551,6 +548,7 @@ func RestoreForest(miscForestFile *os.File, forestFile *os.File) (*Forest, error
 		d := new(diskForestData)
 		d.f = forestFile
 		f.data = d
+		d.cache = newDiskForestCache(24)
 	}
 	f.positionMap = make(map[MiniHash]uint64)
 
--- a/accumulator/forest.go
+++ b/accumulator/forest.go
@@ -111,6 +111,10 @@ func NewForest(forestFile *os.File) *Forest {
 		d := new(diskForestData)
 		d.f = forestFile
 		f.data = d
+		d.cache = diskForestCache{
+			Size: 1 << 16, // 2^16 leaves
+			data: make(map[uint64]Hash),
+		}
 	}
 
 	f.data.resize(1)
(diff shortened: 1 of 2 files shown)


adiabat:
Looks good -- I'm a bit hesitant to merge in this form as it's quite a bit more complex, and if we really need the performance the ramForestData doesn't get that big.  Since diskForestData is so small and this basically rewrites the whole thing, could you instead make this into a 3rd struct that satisfies the interface, something like "cacheForestData"?

I'd guess it would improve speed, but maybe not that much due to writes all still going immediately to disk, and the general pattern being a big mix of reads and writes.  But that seems more complex to cache writes.  But then you could potentially go for a while without any writes, and flush all the changes to disk while doing something else like hashing.  But... yeah that sounds very tricky.

Another similar idea would be to do block-at-a-time reads and writes; get `Modify()` to tell you all the positions it needs to read, and all the positions it will need to write.  It's possible to know this just from the output of the transformation, without knowing any of the actual hashes.  Then you could do a read-everything-at once into ram, and do all the hashing, and then write everything to disk at once.  That might be simpler than write cache, but also needs more changes to several places in `Modify()`

You could definitely go quite far with this, I just wonder how much the disk i/o is really the bottleneck though; maybe something simple like this is enough.  A first benchmark to look at is how much faster genproofs runs with ramForestData vs diskForestData.  If they're similar in speed, caching won't help, but if diskForestData is much slower caching is worth it.

dergoegge:
> could you instead make this into a 3rd struct that satisfies the interface, something like "cacheForestData"?

Good idea, makes sense.

> I'd guess it would improve speed, but maybe not that much due to writes all still going immediately to disk, and the general pattern being a big mix of reads and writes.

As far as i can tell right now it seems to be faster in the beginning since the whole forest fits into the cache but after that there are more disk operations than cache operations, and the speed is similar to no cache at all. So it's sort of the same as using the ram until you don't have enough ram to fit the whole forest and then use the disk.

> A first benchmark to look at is how much faster genproofs runs with ramForestData vs diskForestData.

Good idea, may be the cache isn't even needed (i should have done this before writing the code :D).   I will try to do that later today.

dergoegge:
@adiabat Benchmarks for the first 100k blocks. Seems like the runtimes of ram and disk diverge quite a bit over time, so a cache might be worth it.
I used an external HDD that makes it worse probably.

Time since start
Average read_write durations

dergoegge:
Compared to the cache data. As expected the cache is slower than ram but faster than disk.

Time since start (1)
Average read_write durations (1)

dergoegge:
@adiabat `diskForestData` is still the default forestdata type, so if you want to test this you will have to change `accumulator.NewForest` to use `cacheForestData` (see usage example in the PR description).
There should probably be a cli flag to choose the forestdata type/how much ram should be dedicated to the cache.

adiabat:
Great, can add cli options for forest type for genproofs

PR #149, opened 2020-06-18
.github/: Add an integration test step

dergoegge:


dergoegge:
It's kind of stupid that the download of bitcoind adds ~2min.

kcalvinalvin:
lines 685 and 686 on integration test logs
```
User exit signal received. Exiting...
##[error]blockServer accept error: accept tcp 127.0.0.1:8338: use of closed network connection
```

Are happening because of the 5 min timeout right?

Besides that everything looks good :)

dergoegge:
That "error" is printed when ever genproofs gets killed because the goroutine that accepts connections fails because the `listener` got closed. 
https://github.com/mit-dci/utreexo/blob/dab732bba431021512e2fac7b18316aeb26a28fe/bridgenode/server.go#L54-L58
The script kills genproofs if ibdsim exits without error:
https://github.com/mit-dci/utreexo/blob/dab732bba431021512e2fac7b18316aeb26a28fe/test/csn_bridge.sh#L147
so in this case it's not because of the 5min timeout.

I am not sure why the ci highlights this message because it's not printed to stderr, but i guess it has the word "error" in it :D

kcalvinalvin:
> That "error" is printed when ever genproofs gets killed because the goroutine that accepts connections fails because the `listener` got closed. 
> 
> https://github.com/mit-dci/utreexo/blob/dab732bba431021512e2fac7b18316aeb26a28fe/bridgenode/server.go#L54-L58
> 
> The script kills genproofs if ibdsim exits without error:
> 
> https://github.com/mit-dci/utreexo/blob/dab732bba431021512e2fac7b18316aeb26a28fe/test/csn_bridge.sh#L147
> 
> so in this case it's not because of the 5min timeout.
> 
> 
> 
> I am not sure why the ci highlights this message because it's not printed to stderr, but i guess it has the word "error" in it :D
> 
> 

ok cool so not even an actual error :)

PR #150, opened 2020-07-02
offsetfile: Buffer the leveldb, reader, and writer for offestfile

kcalvinalvin:
Offsetfile building but now with buffers. In my machine, sys and user time is 80% less but the wall clock is only showing 30% less time :/ I guess there's some serious disk io bottleneck going on that *could* be improved with more effort. Just not sure if it's worth the effort.

Benchmark added in offestfile_test.go and also fixes a test in offsetfile_test.go which was broken in master. We could also start to use the tests here...

### DIFF
--- a/bridgenode/offsetfile.go
+++ b/bridgenode/offsetfile.go
@@ -1,6 +1,7 @@
 package bridgenode
 
 import (
+	"bufio"
 	"crypto/sha256"
 	"encoding/binary"
 	"fmt"
@@ -9,7 +10,6 @@ import (
 	"sync"
 
 	"github.com/mit-dci/utreexo/util"
-	"github.com/syndtr/goleveldb/leveldb"
 )
 
 // buildOffsetFile builds an offsetFile which acts as an index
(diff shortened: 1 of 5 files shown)


dergoegge:
I ran the tests. it works and it it's significantly faster!

dergoegge:
also tested with an external HDD, my internal SSD is about 10x faster but before buffering it took for ever on the HDD.

kcalvinalvin:
Merged #151

bridgenode/offsetfile.go
```diff
@@ -48,14 +48,18 @@ func buildOffsetFile(dataDir string, tip util.Hash,
 		}
 	}
 
-	lvdb, err := OpenIndexFile(dataDir)
-	if err != nil {
-		panic(err)
-	}
-	defer lvdb.Close()
+	lvdb := OpenIndexFile(dataDir)
```
adiabat, line 51: Doesn't catch the error...
actually I think I can merge anyway and just put it in

adiabat:
Looks good!
(Well actually doesn't compile but that's a rebase thing, works fine with that)

PR #148, opened 2020-06-16
Provide hooks for wallet operation

adiabat:
splits cmd into server / client, changes paths, other changes.
also scans for addresses for CSN, thought not well yet

PR #151, opened 2020-07-02
Fix errors introduced with #148

dergoegge:
There were some small errors introduced with #148, which this PR fixes.

### DIFF
--- a/bridgenode/genproofs.go
+++ b/bridgenode/genproofs.go
@@ -42,7 +42,7 @@ func BuildProofs(
 		fmt.Printf("not in %s, specify alternate path with -datadir\n.", dataDir)
 		return err
 	}
-	knownTipHeight = 729000
+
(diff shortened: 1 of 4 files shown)


bridgenode/genproofs.go
```diff
@@ -42,7 +42,7 @@ func BuildProofs(
 		fmt.Printf("not in %s, specify alternate path with -datadir\n.", dataDir)
 		return err
 	}
-	knownTipHeight = 729000
```
kcalvinalvin, line 45: @adiabat was this for testing purposes?

adiabat: Oops yeah, didn't want to wait for it to finish (took forever).  Can get rid of that.  Useful one-liner when you don't want to wait to test things though!

adiabat:
Cool - I don't know if the check for closed channel got removed, but that works!

PR #152, opened 2020-07-04
bridgenode: Sync live blocks

dergoegge:
Seems to work but so far i only tested on regtest

- [x] Use the scantxoutset rpc instead of gettxoutsetinfo
- [x] Make the integration test work
- [ ] Handle abandoned chaintips (my testnet node has a couple of those)


dergoegge:
Closing because we won't do live blocks for the first "release" and after that we will integrate the accumulator into btcd, so it's not worth it.

PR #156, opened 2020-07-08
Added Windows instructions

antiquicksort:


adiabat:
Looks good, thanks!

One change I'd request is that the codebase recently changed, and there is no longer a single cmd binary, as it's been split into server and client binaries.  If you change the readme to reflect the current commands (utreexoserver and utreexoclient) it should be good to merge.

kcalvinalvin:
Should also add that you should have ```utreexoserver``` running in order for ```utreexoclient``` to work

PR #120, opened 2020-05-03
utreexo/accumulator: Add documentation for accumulator

kcalvinalvin:
Adds a doc.go for accumulator package and a transformreadme.md for
transform.go file. leafDataReadme.md is edited into leafdatareadme.md

Doesn't change any functionality or code. Just adds/changes documentation

PR #158, opened 2020-07-14
csn: Faster IBD

kcalvinalvin:
In my testing vs master to block #120,000, this branch syncs in ~4m50s vs master of ~8m20s.

This PR depends on changes to btcd code which I have made a PR on at https://github.com/Rjected/btcd/pull/1

### DIFF
--- a/util/utils.go
+++ b/util/utils.go
@@ -71,9 +71,10 @@ func UblockNetworkReader(
 		panic(err)
 	}
 
+	// TODO goroutines for only the Deserialize part might be nice.
+	// Need to sort the blocks though if you're doing that
 	for ; ; curHeight++ {
-		var ub UBlock
-		err = ub.Deserialize(con)
+		block, err := ReadBlock(con)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


util/utils.go
```diff
@@ -71,22 +71,54 @@ func UblockNetworkReader(
...
+
 		var ub UBlock
-		err = ub.Deserialize(con)
+		bReader := bytes.NewReader(block)
+		err = ub.Deserialize(bReader)
 		if err != nil {
 			if err == io.EOF {
 				close(blockChan)
 				break
 			}
```
dergoegge, line 93: This should be moved to the error handling of ReadBlock, then if we get an EOF like in the github action right now it breaks and is done.

kcalvinalvin: Oops. Fix applied :)

adiabat:
Cool, IBD speedups!  I will change things so that `putBlockInPollard` becomes a method on `Csn`, and then give the params to `CheckBlock`

PR #159, opened 2020-07-14
bridgenode: Improve BlockAndRevReader performance

dergoegge:
Problem: the `BlockAndRevReader` makes a ton of syscalls because it reads each block individually from the blk files.
Solution: read multiple blocks from the blk files at once.

pprof results:

Before: 
Screenshot 2020-07-14 at 21 07 20

After:
Screenshot 2020-07-14 at 21 08 04

### DIFF
--- a/bridgenode/rev.go
+++ b/bridgenode/rev.go
@@ -84,6 +84,9 @@ func BlockAndRevReader(
 	}
 }
 
+// GetRawBlocksFromDisk retrives multiple consecutive blocks starting at height `startAt`.
+// `count` is a upper limit for the number of blocks read.
+// Only blocks that are contained in the same blk file are returned.
 func GetRawBlocksFromDisk(startAt int32, count int32, offsetFileName string,
 	blockDir string) (blocks []wire.MsgBlock, revs []RevBlock, err error) {
 	if startAt == 0 {
@@ -114,10 +117,16 @@ func GetRawBlocksFromDisk(startAt int32, count int32, offsetFileName string,
 	revOffsets := make([]uint32, count)
 
 	var datFileNum uint32
+	// minOffset holds the smallest block offset that was seen
 	minOffset := uint32(math.MaxUint32)
+	// maxOffset holds the largest block offset that was seen
 	maxOffset := uint32(0)
+	// minRevOffset holds the smallest rev offset that was seen
 	minRevOffset := uint32(math.MaxUint32)
+	// maxRevOffset holds the largest rev offset that was seen
 	maxRevOffset := uint32(0)
+	// offsetsRead holds the number of blocks < count that are loacated in
+	// blk file with number `datFileNum`
 	offsetsRead := uint32(0)
 	for ; offsetsRead < uint32(count); offsetsRead++ {
 		var datFileNumTmp uint32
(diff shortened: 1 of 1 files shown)


bridgenode/rev.go
```diff
@@ -62,19 +64,166 @@ func BlockAndRevReader(
...
+	if err != nil {
+		return
+	}
+	defer blockFile.Close()
+	blockFileInfo, _ := blockFile.Stat()
+
+	// Read all block data needed for the blocks into memory.
+	// The blocks that are needed are located between the minOffset and
+	// maxOffset+(the size of the block at maxOffset).
+	// 1<<23 = +8MB TODO: whats the max on disk size of a block?
```
kcalvinalvin, line 181: blk.dat files have a limit of 128MB. For us, we're really only gonna be dealing with 128MB blk files since we don't live sync.

I think it'd be better to have a constant of 128MB for the buffer rather than the current method in the code

bridgenode/rev.go
```diff
@@ -62,19 +64,166 @@ func BlockAndRevReader(
...
+	if err != nil {
+		return
+	}
+	defer revFile.Close()
+	revFileInfo, _ := revFile.Stat()
+
+	// Read all rev data needed for the blocks into memory.
+	// The rev data thats needed is located between the minRevOffset and
+	// maxRevOffset+(the size of the rev at maxRevOffset).
+	// 1<<23 = +8MB TODO: whats the max on disk size of a rev?
```
kcalvinalvin, line 199: Rev blocks contain the spend TXOs so the limit is on them. The theoretical limit is 240MB as a Bitcoin block can have a max of 23487 inputs with theoretical max of 10KB per UTXO.

But realistically no one creates UTXOs that big so 128MB here would suffice as well. I don't think there are rev files that are bigger than ~50MB so I think we're safe.

bridgenode/rev.go
```diff
@@ -62,19 +64,166 @@ func BlockAndRevReader(
...
+	_, err = revFile.ReadAt(revData, int64(minRevOffset))
+	if err != nil {
+		return
+	}
+
+	blocks = make([]wire.MsgBlock, offsetsRead)
+	revs = make([]RevBlock, offsetsRead)
+	skip := make([]byte, 8)
+	for i := uint32(0); i < offsetsRead; i++ {
+		blockBuf := bytes.NewBuffer(blockData[offsets[i]-minOffset:])
```
kcalvinalvin, line 210: These aren't really ideal since a new buffer per loop does put some gc pressure. It's small now as the other bottlenecks take so much time so it's fine for now but something to note for the future.

kcalvinalvin:
Tested and things are a lot faster :) I think getting rid of the ```maxoffset``` variables would be good though.

bridgenode/rev.go
dergoegge: done. this is nice because now there is no need for the min/max offsets.

bridgenode/rev.go
dergoegge: i am only using the buffer to have a io.Reader to pass to `MsgBlock.Deserialize`. bytes.Buffer sadly does not support Seek otherwise one buffer would be enough.

adiabat:
Tested, looks good & a lot faster, cool!

PR #160, opened 2020-07-17
watch address and report transactions

adiabat:
This PR adds a command line option to utreexoclient that takes a bech32 address and watches for it as it runs through IBD.

I figure that gives you something to watch while it's doing IBD which is otherwise kind of boring.

It could give a "balance", or list of existing utxos or something.

This seems like enough to show that a wallet could work, without having to write or attach an actual wallet.


adiabat:
OK finally got the test to make a green check mark!  Merging

PR #161, opened 2020-07-20
offsetfile: Ignore blocks that don't have undo data

dergoegge:
fixes the `unexpected EOF` crash at the end of genproofs.


adiabat:
Looks good, thanks! Satoshi's mysteries of rev.dat files :)

PR #163, opened 2020-07-21
csn: Print the amount of satoshis found at the end

dergoegge:


adiabat:
Cool, everyone can try to get a high score!

PR #165, opened 2020-07-21
Various Performance improvements

kcalvinalvin:
Changes were made to the accumulator and the utils

accumulator: More efficient pollard

Two things are changed:

1. Pollard no longer generates goroutines per parentHash in rem2. The
  bottleneck would come from the go scheduler deciding which goroutine to
  run. Resulting in the scheduler block being longer then the execution
  time. This also solves the datarace issue where duplicate hashes would
  be made

2. Gets rid of fmt from grabPos. This is helpful for reducing the gc
  latency as every fmt is a heap allocation.

utils: Restore buffered wire read and writes

This rolls back the changes so that the server and client doesn't do
many small reads, but rather 2 reads/sends per block

### DIFF
--- a/accumulator/hasher.go
+++ b/accumulator/hasher.go
@@ -1,26 +1,9 @@
 package accumulator
 
-import (
-	"sync"
-)
-
 // hashableNode is the data needed to perform a hash
 type hashableNode struct {
 	sib, dest *polNode
 	position  uint64 // doesn't really need to be there, but convenient for debugging
-}
-
-// this should work, right?  like the pointeryness?  Because swapnodes doesn't
-// change pointers.  Well it changes niece pointers, but the data itself changes
-// so the data is "there" in a static structure so I can use pointers to it
-// and it won't move around.  hopefully.
-
-func (n *hashableNode) run(wg *sync.WaitGroup) {
-	// fmt.Printf("hasher about to replace %x\n", n.dest.data[:4])
-	n.dest.data = n.sib.auntOp()
-	// fmt.Printf("hasher finished %x %x -> %x\n",
-	// n.sib.niece[0].data[:4], n.sib.niece[1].data[:4], n.dest.data[:4])
-	wg.Done()
 }
 
 type hashNpos struct {
--- a/accumulator/pollard.go
+++ b/accumulator/pollard.go
@@ -2,8 +2,17 @@ package accumulator
 
 import (
 	"fmt"
-	"sync"
 )
+
+const (
+	ErrorNotEnoughTrees uint32 = iota
+	ErrorNoPollardNode
+)
+
+var ErrorStrings = map[uint32]error{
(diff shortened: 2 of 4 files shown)


accumulator/hasher.go
```diff
@@ -15,12 +11,11 @@ type hashableNode struct {
 // so the data is "there" in a static structure so I can use pointers to it
 // and it won't move around.  hopefully.
 
-func (n *hashableNode) run(wg *sync.WaitGroup) {
```
ysangkok, line 18: since this is now a one-liner used in only one location, wouldn't it be better to just inline it? since there is no longer even a go routine, i think the name "run" doesn't add any value at all.

kcalvinalvin: Hmm yeah good call I thought maybe the compiler would inline it but it doesn't. Will change

accumulator/pollard.go
```diff
@@ -333,37 +339,39 @@ func (p *Pollard) swapNodes(s arrow, row uint8) (*hashableNode, error) {
...
+		// grab the sibling of lr
+		lrSib := lr ^ 1
 		if h == 0 { // if at bottom, done
 			hn.dest = nsib // this is kind of confusing eh?
 			hn.sib = n     // but yeah, switch siblingness
-			n, nsib = n.niece[lr^1], n.niece[lr]
+			n, nsib = n.niece[lrSib], n.niece[lr]
 			if nsib == nil || n == nil {
-				// fmt.Printf("gave up ")
 				return // give up and don't make hashable node
```
ysangkok, line 360: while you're cleaning this up, maybe this comment could be improved. what does "don't make hashable node" mean @adiabat ? it was already made

accumulator/pollard.go
```diff
@@ -155,7 +164,7 @@ func (p *Pollard) rem2(dels []uint64) error {
 
 	// get all the swaps, then apply them all
 	swaprows := remTrans2(dels, p.numLeaves, ph)
-	wg := new(sync.WaitGroup)
+	//wg := new(sync.WaitGroup)
```
ysangkok, line 167: i don't see the value in having this dead code laying around

accumulator/pollard.go
kcalvinalvin: Made improvement to the comment.

accumulator/pollard.go
kcalvinalvin: Removed

ysangkok:
thanks for following up and cleaning things up @kcalvinalvin

accumulator/pollard.go
adiabat: We could just get rid of the whole `if nsib == nil || n == nil {` lines since they don't do anything.  Once you get to line 358 it's done and returning no matter what.
So the comment is wrong...

kcalvinalvin: Removed the branch

kcalvinalvin:
Client reader broke and I would have to rewrite the code as I lost the commit. Closing

PR #166, opened 2020-07-21
csn: Save utxos between runs

dergoegge:


adiabat:
Cool it's almost a full wallet now :)

PR #167, opened 2020-07-21
accumulator: More efficient pollard

kcalvinalvin:
Two things are changed:

1. Pollard no longer generates goroutines per parentHash in rem2. The
bottleneck would come from the go scheduler deciding which goroutine to
run. Resulting in the scheduler block being longer then the execution
time. This also solves the datarace issue where duplicate hashes would
be made

2. Gets rid of fmt from grabPos. This is helpful for reducing the gc
latency as every fmt is a heap allocation.

### DIFF
--- a/accumulator/hasher.go
+++ b/accumulator/hasher.go
@@ -1,26 +1,9 @@
 package accumulator
 
-import (
-	"sync"
-)
-
 // hashableNode is the data needed to perform a hash
 type hashableNode struct {
 	sib, dest *polNode
 	position  uint64 // doesn't really need to be there, but convenient for debugging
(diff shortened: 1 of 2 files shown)


adiabat:
Looks good 

As discussed today, `grabPos` is a bottleneck.  This helps, and there's also a larger speedup which would need a big change to `grabPos` - right now it descends from the root each time.  If you request 100 leaves, it's going to start and the root and descend the whole length each time which is very inefficient.  A `grabMultiPos` function could take in a slice of positions and return a slice of the results and should be way more efficient.
So that could be a TODO for later.

PR #168, opened 2020-07-21
Make server listen publicly and client can connect to specified host

adiabat:
note that this makes the server listen on the network interface!
so it's an actual server now, whatch out


adiabat:
This works but is slow - you need a round trip per block, and can't request block n+1 until you get block n.  This hasn't changed, but I'm just realizing it since I never actually tried it other than localhost till now.

So that's a definite TODO to fix.

PR #170, opened 2020-07-21
enable backwards validation

adiabat:
could be fun!  Start at the tip, and work back to the genesis block

### DIFF
--- a/bridgenode/server.go
+++ b/bridgenode/server.go
@@ -66,8 +66,15 @@ func serveBlocksWorker(c net.Conn, endHeight int32, blockDir string) {
 	fmt.Printf("start serving %s\n", c.RemoteAddr().String())
 	var curHeight int32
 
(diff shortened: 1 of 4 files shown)


adiabat:
The "easy" way would be for every udata to also include a rootSet of what the starting state of the accumulator should be.  This is redundant for clients going forward since they already have it, and I think it's also redundant for clients going backwards.  But -- it's simple to verify for the backwards clients.

It seems there's also a fancier version where the backwards clients figure this out without any hints from the server, but I need to figure out how that works.  For example, in blocks with no proofs (empty blocks which don't consume any utxos but just reward the miner),  you need to defer linking up the rootSets, as you know what to add (the miner utxo) but there's nothing to remove and no proofs to verify.

dergoegge:
We are already distingushing between backwards and forwards clients so sending the rootsets to backwards clients only would also work.

PR #173, opened 2020-07-22
Various linting and spelling nits

MDrollette:
Cleaning up some linter warnings and typos just to get familiar with the project.


accumulator/types.go
```diff
@@ -39,7 +39,7 @@ type node struct {
 	Val Hash
 }
 
-// LeafTXOs have a hash and a expiry date (block when that utxo gets used)
+// Leaf TXOs have a hash and a expiry date (block when that utxo gets used)
```
kcalvinalvin, line 42: This is a stale comment. Since this is a cleanup PR, could you change this to ```Leaf contains a...```

kcalvinalvin:
Looks good! Could you squash the commits into one commit

adiabat:
Looks great, thanks!  One change request:
The 
` for a, b := range someSlice {`
syntax -- I've had trouble, and someone else working on this repo as well (forget who) with the ambiguity of `for a:= range someSlice`.  Especially when ranging over a slice of ints!  So I've been sticking to `a, _ :=` and `_, b :=` for all uses of range so it's more explicit.  I guess that may not be the default style in go but it has helped me and others a few times so I'd like to keep those.

PR #176, opened 2020-07-27
accumulator/pollard: Speedups for pollard

kcalvinalvin:
Things done:

1. Different implementation of #3 where the already known hashes of parents were matched, saving sha256 hashes. Instead of going through all the nodes to the path of the leaf being verified, in this implementation just looks for the polNode at the parent position. New method on pollard ```readPos``` was needed as reading from ```grabPos``` created nil hashable nodes.

2. Various comments and small code fixes.

In my testing to block 600,000 on testnet3, the speedups were around 5%. However, when tested on blocks with lower height, the speedups were even less. So more hashing goes on as the tree gets bigger, leading to bigger savings. The matched hashes to the tip of the default server was 160,716,082 hashes and the unmatched hashes were 55,700,940 hashes. So this is around 75% of potential sha computation saved.

### DIFF
--- a/accumulator/batchproof.go
+++ b/accumulator/batchproof.go
@@ -20,7 +20,6 @@ func (bp *BatchProof) ToBytes() []byte {
 	var buf bytes.Buffer
 
 	// first write the number of targets (4 byte uint32)
-
 	numTargets := uint32(len(bp.Targets))
 	if numTargets == 0 {
 		return nil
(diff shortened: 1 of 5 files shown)


accumulator/pollardproof.go
```diff
@@ -37,8 +34,7 @@ func (p *Pollard) IngestBatchProof(bp BatchProof) error {
 			if node.niece[lr] == nil {
 				node.niece[lr] = new(polNode)
 				node.niece[lr].data = proofMap[pos]
-				// fmt.Printf("------wrote %x at %d\n", proofMap[pos], pos)
-				if node.niece[lr].data == empty {
+				if node.niece[lr].data == (Hash{}) {
```
dergoegge, line 37: would this not allocate 32 bytes for every comparison? or is the go compiler smart enough to avoid that?

accumulator/batchproof.go
```diff
@@ -223,25 +231,24 @@ func verifyBatchProof(
...
 	}
-	proofCopy := make([]Hash, len(bp.Proof))
-	copy(proofCopy, bp.Proof)
 	proofTree := make(map[uint64]Hash)
 
+	// If there is nothing to reconstruct, return empty map
 	if len(bp.Targets) == 0 {
 		return proofTree, nil
 	}
+	proofCopy := bp.Proof // back up proof
```
dergoegge, line 248: nit: maybe call this just `proof`, `targets` is also not called `targetsCopy`.

accumulator/pollardproof.go
```diff
@@ -79,3 +75,139 @@ func (p *Pollard) IngestBatchProof(bp BatchProof) error {
...
+					left, proofmap[left], right, proofmap[right], parentPos)
+			}
+			var parhash Hash
+			n, _, _, err := p.readPos(parentPos)
+			if err != nil {
+				if err == ErrorStrings[ErrorNoPollardNode] {
+					panic(err)
+				} else {
+					panic(err)
+				}
```
dergoegge, line 182: i guess you had the if statement here for testing, could remove it and just panic

accumulator/pollardproof.go
kcalvinalvin: Looks like the compiler is not smart enough to do so. I'll revert this

accumulator/batchproof.go
kcalvinalvin: Sure. This really shouldn't even be a method though if there's gonna be a copy... Something to do later down the road.

accumulator/pollardproof.go
kcalvinalvin: will fix

adiabat:
Tested, looks good.  Could have even more speedups combined with #177, or further TTL / caching after that PR.

PR #177, opened 2020-07-28
Use UTXO time-to-live values

dergoegge:


bridgenode/server.go
```diff
@@ -34,7 +37,7 @@ func blockServer(endHeight int32, dataDir string, haltRequest, haltAccept chan b
 			close(cons)
 			return
 		case con := <-cons:
-			go serveBlocksWorker(con, endHeight, dataDir)
+			go serveBlocksWorker(con, endHeight, dataDir, lvdb)
```
adiabat, line 40: Right, now the server needs access to the db to get the TTL values.  OK for now but we should fix this, as that could be a bottleneck for serving to lots of clients.
Probably should have the DB for only temporary use and write the TTL somewhere in the flat files once it's known.

adiabat:
This looks good.  The changes in this PR might not affect IBD times much, but it's set up for further changes which will.
Some next things to do for TTL / caching would be:

- Get TTL values into the proof flat file instead of leaving them in a levelDB folder.  Once the TTL value for a utxo is known we can write it into the flat file.  We can allocate 4 bytes per UTXO and then keep track of the byte offset of the utxo TTL value in the levelDB entry, then write the TTL value to that offset and remove the levelDB entry once the UTXO is spent.  This would limit the TTL db to only the size of the current UTXO set, and then let the server not have to do many DB reads.

- Change wire protocol for the server / client so it doesn't send the whole proof.  That's harder / more annoying.  I think the simplest way (which is still not all that simple) is for the server to first send the positions, without any hashes (the `Targets []uint64` part of batchProof without the `Proof   []Hash` part) and then the client replies with a list of hash positions required. 
A more network-efficient way would be for the server to maintain a bunch of state for each client, and remember what positions in the client's pollard are filled / empty, and then send only what's needed.  That's a lot of work for the server for each client though and feels more easily DoS-able.

kcalvinalvin:
@adiabat hrmmm wait wouldn't a ttl for a utxo that's spent after the ttl is written to a flat file become stale/incorrect?

kcalvinalvin:
Also, could we start squashing the commits so that it's one commit per feature?

Reason-ale:

In a case like such (and this has happened during the binary release), it makes hard to have a clean git history:

1. Bob makes changes to his local repo and is ready to push.
2. Alice's PR gets merged. This PR has many commits but is just for a feature.
3. Bob merges Alice's tree
4. Bob needs to make a change to one of his commits. This can only be done by rewriting the history (commits in the merge) or adding an additional "fix" commit.

If Alice's PR was only a commit, then it's a lot easier to handle for Bob.

For this reason, all major repos that I am aware of is doing this, including Bitcoin Core.

This is a simple change but is really helpful in a open source environment with many contributors :)

adiabat:
> @adiabat hrmmm wait wouldn't a ttl for a utxo that's spent after the ttl is written to a flat file become stale/incorrect?

You can't write any TTL for an output until it's spent, so you just but a placeholder 4 bytes there initially (0 or ffff or whatever)
Then when the output gets spent you overwrite the placeholder TTL with the actual TTL, and remove the entry from levelDB.  

For squashing, should I do the squash and merge option, or should people squash first and then use the regular merge pull request?  (or maybe it does the same thing)

kcalvinalvin:
hmm right. Not quite sure how expensive the updating the ttls would be (I guess not that much since leveldb isn't that much faster anyways on one-off reads?). I don't think we update the ttls right now either so that's also a todo.

I think letting the author squash is better since if the maintainer squash-merges, they'll be (can be) rewriting history.

bridgenode/server.go
dergoegge: I think if the leveldb has enough cache performance shouldn't be too bad, but yeah having the ttls in flat files would be nicer.

kcalvinalvin:
Tested! Looks good

adiabat:
TODO: Put TTL data into proof flat file so this doesn't slow things down

PR #178, opened 2020-07-30
extractTwins example in comment

ysangkok:
I think an example could provide some intuition on what this function is doing, since I didn't get it before I started working through the example I was given on IRC. I would understand if you think examples are not the best way to explain it, we could also put them somewhere else if you feel it is too verbose. But I submit it anyway, since I think it could help people randomly looking through the source, and I think it is better to be too verbose than too cryptic.

### DIFF
--- a/accumulator/utils.go
+++ b/accumulator/utils.go
@@ -10,6 +10,16 @@ var verbose = false
 
 // takes a slice of dels, removes the twins (in place) and returns a slice
 // of parents of twins
+//
+// Example with 15-leaf tree in printout.txt line 21:
+// If deleting the nodes 4 through 8, the function would be called with
+// nodes=[4,5,6,7,8] and row=4. (note that this example different from
+// others, since it has been constructed to have both parents and dels)
+// It would return (parents=[18,19], dels=[8]) because with the given
+// amount of rows (row=4), only 8 is not a "twin" (it has no siblings
+// to be deleted, note that 7 is in a different tree than 8).
+// [18, 19] are each parents of twins, since their children are nodes
+// 4-7 which are to be deleted.
 func extractTwins(nodes []uint64, row uint8) (parents, dels []uint64) {
 	for i := 0; i < len(nodes); i++ {
 		if i+1 < len(nodes) && nodes[i]|1 == nodes[i+1] {


accumulator/utils.go
```diff
@@ -10,6 +10,15 @@ var verbose = false
 
 // takes a slice of dels, removes the twins (in place) and returns a slice
 // of parents of twins
+//
+// Example with 15-leaf tree in printout.txt line 21:
+// If deleting the nodes 4 through 9, the function would be called with
+// nodes=[4,5,6,7,8] and row=4.
```
kcalvinalvin, line 16: row would be 3 since rows are counted from 0. So the 15-leaf tree in printout.txt would be 3. This can be seen with pollard.rows().

accumulator/utils.go
```diff
@@ -10,6 +10,15 @@ var verbose = false
 
 // takes a slice of dels, removes the twins (in place) and returns a slice
 // of parents of twins
+//
+// Example with 15-leaf tree in printout.txt line 21:
+// If deleting the nodes 4 through 9, the function would be called with
```
kcalvinalvin, line 15: shouldn't it be 4 through 8?

kcalvinalvin:
Verbose is good! I think all of us agree that the code is too cryptic. Left some comments on the PR

accumulator/utils.go
ysangkok: I think row is forestRows here. forestRows of the tree I mentioned is 4. Do you agree?

ysangkok: This is what I tried:

```
diff --git a/accumulator/forest_test.go b/accumulator/forest_test.go
index a1595c3..f650ef1 100644
--- a/accumulator/forest_test.go
+++ b/accumulator/forest_test.go
@@ -53,7 +53,7 @@ func TestForestAddDel(t *testing.T) {
 
 func TestForestFixed(t *testing.T) {
        f := NewForest(nil, false)
-       numadds := 5
+       numadds := 15
        numdels := 3
        adds := make([]Leaf, numadds)
        dels := make([]uint64, numdels)
@@ -71,6 +71,7 @@ func TestForestFixed(t *testing.T) {
        if err != nil {
                t.Fatal(err)
        }
+       panic(f.rows)
        fmt.Print(f.ToString())
        fmt.Print(f.PrintPositionMap())
        _, err = f.Modify(nil, dels)
```

It paniced with 4.

I am a bit confused. What are you proposing?

accumulator/utils.go
ysangkok: I changed it to be 4 though 9 everywhere to be consistent with the other PR.

kcalvinalvin: That works but if you are deleting leaves 4-9, then 8 has a sibling (9) so the below comment would be off where it states:

```only 8 is not a "twin" (it has no siblings to be deleted, note that 7 is in a different tree)```

Above you mention that 20 is returned as a parent. 20 would not be returned if 8 didn't have a sibling.

For deletion of nodes [4, 5, 6, 7, 8, 9], the returning values would be ```parents = [18, 19, 20]``` and ```dels=[]```

File: accumulator/utils.go (hunk 1 above)
Comment (kcalvinalvin) [reply accumulator/utils.go]: Ah oops yeah you're right the height is 4. I forgot that the allocated rows is the one that gets returned.

But for future reference I should mention that the height is 4 not because you count from 1 but because the allocated tree always has 2^n leaves. The tree at line 21 in printout.txt is really this:

```
em
|-------------------------------\
28                              em
|---------------\               |---------------\
24              25              26              em
|-------\       |-------\       |-------\       |-------\
16      17      18      19      20      21      22      em
|---\   |---\   |---\   |---\   |---\   |---\   |---\   |---\
00  01  02  03  04  05  06  07  08  09  10  11  12  13  14  em
```

The tree just didn't list the empty nodes.

ysangkok:
@kcalvinalvin Ok, I reverted to deleting 4-8 since that gives an opportunity to have examples in both of the returned lists.

adiabat:
Cool, this works.

After merge I want to change a bit though; 

```extractTwins(nodes []uint64, row uint8) (parents, dels []uint64) {```

should be 
```extractTwins(nodes []uint64, forestRows uint8) (parents, dels []uint64) {```

to be consistent with all the other functions.

PR #179, opened 2020-07-30
makeCollapse example in comment

ysangkok:
I added some parameter docs, and I tried adding an example, but it is mostly just a run through the function with example data for the first branch. Since it only covers one branch, I feel that the docs could still be improved. But I don't understand it all, so I submit what I have. I'd appreciate if someone could tell me whether they think this comment could improve comprehension. Maybe I should remove the example?

### DIFF
--- a/accumulator/transform.go
+++ b/accumulator/transform.go
@@ -64,6 +64,48 @@ func remTrans2(dels []uint64, numLeaves uint64, forestRows uint8) [][]arrow {
 	return swaps
 }
 
+// makeCollapse determines whether a collapse should take place given a list
+// of nodes to delete and misc info. If the collapse is present, it consists
+// of an single arrow (a source index and a destination index).
+//
+// A collapse is a deferred swap, that swaps a root or the sibling of a
+// leftover deletion to its new position in the modified tree. A collapse can
+// not directly be appended to the regular swaps because the collapse
+// destination changes with upper row swaps. swapCollapses turns the collapse
+// into a regular swap by computing a new destination position so that after
+// upper row swaps the node is located at the collapse destination.
+//
+// delRemains must be true if dels has an odd length.
+// r is the row number to generate a collapse for, 0 is the leaf level.
+// rootPresent must be true if there is a tree root in the row r.
+// nextNumLeaves is the final amount of leaves, after the whole deletion
+// process. Note that if r!=0, some dels may already have happened, the slice
+// will be shorter.
+//
+// Example using 15-leaf tree in printout.txt line 21: (forestRows=4, numLeaves=15)
+//
+// Let's say we are deleting nodes 4 through 9 (6 nodes, so delRemains=False).
+//
+// For this example, let's take the call to this function
+// that happens when r=0 (bottom row).
+//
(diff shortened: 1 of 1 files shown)


accumulator/transform.go
```diff
@@ -64,6 +64,42 @@ func remTrans2(dels []uint64, numLeaves uint64, forestRows uint8) [][]arrow {
...
+// delRemains must be true if dels has an odd length.
+// r is the row number to generate a collapse for, 0 is the leaf level.
+// rootPresent must be true if there is a tree root in the row r.
+// nextNumLeaves is the final amount of leaves, after the whole deletion
+// process. Note that if r!=0, some dels may already have happened, the slice
+// will be shorter.
+//
+// Example using 15-leaf tree in printout.txt line 21: (forestRows=4, numLeaves=15)
+//
+// Let's say we are deleting nodes 4 through 9 (6 nodes, so delRemains=True).
```
kcalvinalvin, line 82: delRemains is to grab the leftover 1 node in the tree, so if we're deleting 6 leaves, ```makeCollapse``` should match case default since there is no root in the tree and no delRemains.

Transform is outlined in 7.1 Batch deletion process in the utreexo paper for reference.

kcalvinalvin:
I think the comment is good but the example is off.

accumulator/transform.go
ysangkok: what do you mean "there is no root"? there is a root on row 0, which is the one the example uses. the root is 14.

kcalvinalvin: Ah right yeah there is a root. Oops my bad

accumulator/transform.go
```diff
@@ -64,6 +64,43 @@ func remTrans2(dels []uint64, numLeaves uint64, forestRows uint8) [][]arrow {
 	return swaps
 }
 
+// makeCollapse determines whether a collapse should take place given a list
+// of nodes to delete and misc info. If the collapse is present, it consists
+// of an single arrow (a source index and a destination index).
+// A collapse is a deferred swap, which is later made into an actual swap in
```
dergoegge, line 70: i think explaining what a collapse is in more detail would be nice. the paper also makes no mention of them. Maybe something like: "A collapse is a deferred swap, that swaps a root or the sibling of a leftover deletion to its new position in the modified tree. A collapse can not directly be appended to the regular swaps because the collapse destination changes with upper row swaps. swapCollapses turns the collapse into a regular swap by computing a new destination position so that after upper row swaps the node is located at the collapse destination."

kcalvinalvin: Just a note but a "collapse" is called "Root" in the paper. It's outlined in section 7.4

dergoegge: hm not entirely, the `!delRemains && rootPresent` case and `swapCollapses` are not in the paper but maybe those are just implementation details

ysangkok: Ok, I adopted your comment verbatim.

ysangkok:
Ok, I think I addressed everything. I did not insert references to the paper because I feel like the paper may be too different. Maybe it would be good to have a separate document explaining the rough equivalences.

PR #180, opened 2020-08-04
pollard: Add method that returns needed subset of proof positions

dergoegge:


kcalvinalvin:
This is the reply to the ping on irc for

```
07:52 dergoegge: kcalvinalvin: in my testcase in #188, i tell the pollard to remember all leaves above position 8. it does remember all leaves >=10 as expected (readPos returns the polNodes), but for position 9 the polNode is nil and the sibling (8) has a reference to 9 via niece[0]. would it be up to the caller of readPos to handle this or should readPos return 8.niece[0] for 9?
```

(I should've really left a comment on readPos) But the ```readPos()``` method was needed because the functionality needed was *just reading* the pollard. ```readPos()``` does not change the pollard state in any way and just returns whatever it find in the pollard. It will return nil nodes if there is a nil. ```grabPos()``` will attach an empty polNode if the sibling is nil.

I haven't checked the rest of the code in this PR but I've added ```fmt.Println(p.ToString())``` after the modify to visualize the pollard. The result that I get is

```
|---------------------------------------------------------------\
28:9cb3
|-------------------------------\                               |-------------------------------\
                                                                26:e2b5
|---------------\               |---------------\               |---------------\               |---------------\
                                                                20:6886         21:c49b         22:fc5b
|-------\       |-------\       |-------\       |-------\       |-------\       |-------\       |-------\       |-------\
                                                                08:0008         10:000a 11:000b 12:000c 13:000d 14:000e
```

I think 9 is nil for the pollard in question, so ```readPos()``` is doing its job.

dergoegge:
> I think 9 is nil for the pollard in question, so readPos() is doing its job.

if you print `readPos(9)` you can see that n itself is nil but nsib (8) is not nil and still knows 9 via niece[0].
This has todo with the way the pollard remembers nodes. it's weird that 21 does not know 9 but 8 does...

the question is if readPos should return nil even though 8 still knows 9?
the same goes for p.ToString()
or is adding/remembering broken?

kcalvinalvin:
Oh yeah I think that just has to do with the weirdness of the pollard remembering function making a node point at itself with the code here https://github.com/mit-dci/utreexo/blob/f84fc26c9e1dd360a3bdef828067a1226ccfd33e/accumulator/pollard.go#L105

The code here is the only bit of code that handles the pruning.
https://github.com/mit-dci/utreexo/blob/f84fc26c9e1dd360a3bdef828067a1226ccfd33e/accumulator/pollard.go#L124

Which that code calls this bit of code
https://github.com/mit-dci/utreexo/blob/f84fc26c9e1dd360a3bdef828067a1226ccfd33e/accumulator/pollardutil.go#L49-L51

So yeah... I think it is a bad idea. I think the code handling remember should be handling the error you found.

adiabat:
I know it's a draft so if these things change then never mind-- just wondering: 
Why change roots from []polNode to []*polNode?  Does that help anything?

Not specific to this PR, but some of the weird stuff we're dealing with here is because of how  the very top and very bottom of the forest things don't quite fit.  `niece[]` for a root is not actually a niece, since roots have no siblings.  `niece[]` for a leaf also a niece, since leaves have no children.  So for a root, `niece[0]` really means `child[0]`, and for a leaf `niece[0]` means "memorable".  Having a pointer to an interface with 3 different types (root, node, leaf) would make it easier to understand but also use 25% more RAM... so still trying to find a cleaner / better way to represent these things.

dergoegge:
> Why change roots from []polNode to []*polNode? Does that help anything?

It just felt a little cleaner to me for 2 reasons:
* no copying the polNode when adding leaves (polNode.data is not copied, so it's not that bad)
* when adding a leaf that should be remembered and that leaf becomes the height 0 root then that root points to a copy of itself and not actually itself.

it's more of a nitpick change than a real improvement, i can revert it if you want. is there a good reason to not use pointers here?

adiabat:
OK yeah if you think it's easier that works.  I guess it's a bit more consistent since it's `*polNode` in most other places.

Another way we could have the "remember" pointer is just to point to *any* node, so we can specify a "memorable" node with no data that every memorable leaf points to.  That might be easier to deal with?  Not sure it makes any difference but just an idea.

dergoegge:
This PR turned into a ground for experiments. I will split up some of the changes into separate PRs.

PR #185, opened 2020-08-18
accumulator: Add CacheSim type that can simulate pollard modification

dergoegge:
was part of my experiments in #180.
Simulation of pollard modification will be helpful to derive partial proofs without adding extra round trips for every block during IBD.


accumulator/pollardproof.go
```diff
@@ -208,3 +209,63 @@ func (p *Pollard) verifyBatchProof(
...
+}
+
+// Simulate simulates one modification to the pollard.
+// Takes a slice of target positions and a slice of bools representing new leaves (true means "cache the leaf").
+// Returns the proof positions that are needed to prove the targets (including the unkown targets).
+func (c *CacheSim) Simulate(targets []uint64, adds []bool) []uint64 {
+	// Figure out which targets are known/cached and which aren't.
+	var knownTargets, unknownTargets []uint64
+	for _, target := range targets {
+		if c.cached.Bit(int(target)) > 0 {
```
kcalvinalvin, line 236: Nit since it'll be eons until we get to 1 << 63 leaves but this could error out right?

https://play.golang.org/p/wo0-tE8s1KN

accumulator/pollardproof.go
```diff
@@ -208,3 +209,63 @@ func (p *Pollard) verifyBatchProof(
...
+}
+
+// NewCacheSimulator initialises a new cache simulator.
+func NewCacheSimulator(numLeaves uint64) *CacheSim {
+	return &CacheSim{cached: big.NewInt(0), numLeaves: numLeaves}
+}
+
+// Simulate simulates one modification to the pollard.
+// Takes a slice of target positions and a slice of bools representing new leaves (true means "cache the leaf").
+// Returns the proof positions that are needed to prove the targets (including the unkown targets).
```
kcalvinalvin, line 231: Another nit but there's a spelling error on unkown

accumulator/utils.go
```diff
@@ -9,6 +9,85 @@ import (
 // verbose is a global const to get lots of printfs for debugging
 var verbose = false
 
+// ProofPositions returns the positions that are needed to prove that the targets exist.
+func ProofPositions(targets []uint64, numLeaves uint64, forestRows uint8) ([]uint64, []uint64) {
+	// the proofPositions needed without caching.
+	var proofPositions, computedPositions []uint64
+	for row := uint8(0); row < forestRows; row++ {
+		computedPositions = append(computedPositions, targets...)
+		if numLeaves&(1<<row) > 0 && len(targets) > 0 &&
```
kcalvinalvin, line 18: ```numLeaves&(1<<row) > 0``` is checking to see if a root exists right? Comment or maybe even splitting this up to make something like ```rootExists := numLeaves&(1<<row) > 0``` could be better for readability.

kcalvinalvin (commented):
In general, I think the concept is good but do you have tests that you used on this code? Proof code is consensus critical and we've caught one consensus bug here before. I think some tests here are needed.

dergoegge:
> I think some tests here are needed.

For sure, I haven't really tested this yet. I will mark this as a draft for now and add some tests.

accumulator/utils.go
dergoegge: we could probably do this in a lot of places or maybe we create a function that checks for roots that we can call in various places, pretty sure the go compiler would inline simple functions like that.

ysangkok: wasn't there some example very recently where Go didn't inline a really simple function? probably something i read from @kcalvinalvin

accumulator/pollardproof.go
ysangkok: @kcalvinalvin are you saying it will be the int casting that overflows? in that case would it be a problem on 32-bit platforms? are those supposed to be supported?

dergoegge: on a 64-bit system it would cast to a signed int64 and `target` is unsigned, so the casted int could be negative

dergoegge: i don't know what happens on a 32-bit system with all the uint64s we are using 😄

accumulator/utils.go
dergoegge: it should inline all functions that meet these [criteria](https://github.com/golang/go/wiki/CompilerOptimizations#function-inlining).
we can also verify that it inlines them using `-gcflags=-m` when building.

dergoegge:
While writing the unit test i found that this does not work as it currently stands or at least it's not optimal.
Not optimal here means that the simulator claims that you need node x but x is cached and thus wouldn't actually be needed. 

It still reduces the proof sizes in a lot of cases though so maybe it's ok for now. Example for targets A and B: if A is cached and B is not but the proof paths of A and B overlap then the combined proof size for A and B can reduced because the overlap of A can be retrieved from cache. This is similar to how you can batch proofs for multiple targets.

It would probably be tricky to make this optimal... one way would be to keep all nodes in the bit set and not just the leaves but then you would have to do the entire remove transformation which could get tricky because it's a bit set and not a data structure like the pollard with pointers. (also the bit set would be 2x in size)

PR #188, opened 2020-08-21
refactor: Use slices instead of maps

dergoegge:
was part of my experiments in #180.
Refactor `ProveBatch`, `verifyBatchProof` and `IngestBatchProof` to use slices.
Preallocation of slices also helps with GC pressure. 
also includes the `ProofPositions` function which is also part of #185, kinda stupid but otherwise this doesn't build.

### DIFF
--- a/accumulator/batchproof.go
+++ b/accumulator/batchproof.go
@@ -112,122 +112,145 @@ func FromBytesBatchProof(b []byte) (BatchProof, error) {
 // TODO OH WAIT -- this is not how to to it!  Don't hash all the way up to the
 // roots to verify -- just hash up to any populated node!  Saves a ton of CPU!
 
-// verifyBatchProof takes a block proof and reconstructs / verifies it.
-// takes a blockproof to verify, and the known correct roots to check against.
-// also takes the number of leaves and forest rows (those are redundant
-// if we don't do weird stuff with overly-high forests, which we might)
-// it returns a bool of whether the proof worked, and a map of the sparse
-// forest in the blockproof
-func verifyBatchProof(
-	bp BatchProof, roots []Hash,
-	numLeaves uint64, rows uint8) (bool, map[uint64]Hash) {
-
-	// if nothing to prove, it worked
+// verifyBatchProof verifies a batchproof by checking against the set of known correct roots.
+// Takes a BatchProof, the accumulator roots, and the number of leaves in the forest.
+// Returns wether or not the proof verified correctly, the partial proof tree,
+// and the subset of roots that was computed.
+func verifyBatchProof(bp BatchProof, roots []Hash, numLeaves uint64,
+	// cached should be a function that fetches nodes from the pollard and indicates whether they
+	// exist or not, this is only useful for the pollard and nil should be passed for the forest.
+	cached func(pos uint64) (bool, Hash)) (bool, [][3]node, []node) {
 	if len(bp.Targets) == 0 {
-		return true, nil
+		return true, nil, nil
 	}
 
-	// Construct a map with positions to hashes
-	proofmap, err := bp.Reconstruct(numLeaves, rows)
-	if err != nil {
-		fmt.Printf("VerifyBlockProof Reconstruct ERROR %s\n", err.Error())
-		return false, proofmap
+	if cached == nil {
+		cached = func(_ uint64) (bool, Hash) { return false, empty }
 	}
 
-	rootPositions, rootRows := getRootsReverse(numLeaves, rows)
-
-	// partial forest is built, go through and hash everything to make sure
-	// you get the right roots
-
-	tagRow := bp.Targets
-	nextRow := []uint64{}
-	sortUint64s(tagRow) // probably don't need to sort
-
-	// TODO it's ugly that I keep treating the 0-row as a special case,
-	// and has led to a number of bugs.  It *is* special in a way, in that
-	// the bottom row is the only thing you actually prove and add/delete,
-	// but it'd be nice if it could all be treated uniformly.
-
-	if verbose {
-		fmt.Printf("tagrow len %d\n", len(tagRow))
-	}
-
-	var left, right uint64
-
-	// iterate through rows
-	for row := uint8(0); row <= rows; row++ {
-		// iterate through tagged positions in this row
-		for len(tagRow) > 0 {
-			// Efficiency gains here. If there are two or more things to verify,
-			// check if the next thing to verify is the sibling of the current leaf
-			// we're on. Siblingness can be checked with bitwise XOR but since targets are
-			// sorted, we can do bitwise OR instead.
-			if len(tagRow) > 1 && tagRow[0]|1 == tagRow[1] {
-				left = tagRow[0]
-				right = tagRow[1]
-				tagRow = tagRow[2:]
-			} else { // if not only use one tagged position
-				right = tagRow[0] | 1
-				left = right ^ 1
-				tagRow = tagRow[1:]
-			}
-
-			if verbose {
-				fmt.Printf("left %d rootPoss %d\n", left, rootPositions[0])
-			}
-			// check for roots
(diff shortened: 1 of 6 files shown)


accumulator/batchproof.go
```diff
@@ -112,122 +112,134 @@ func FromBytesBatchProof(b []byte) (BatchProof, error) {
...
+	for len(targetNodes) > 0 {
+		var target, proof node
+		target = targetNodes[0]
+		if len(proofPositions) > 0 && target.Pos^1 == proofPositions[0] {
+			// target has a sibling in the proof positions, fetch proof
+			proof = node{Pos: proofPositions[0], Val: bp.Proof[0]}
+			proofPositions = proofPositions[1:]
+			bp.Proof = bp.Proof[1:]
+			targetNodes = targetNodes[1:]
+			goto hash
```
kcalvinalvin, line 191: Could be abstracted into a separate function instead of a goto?

accumulator/batchproof.go
```diff
@@ -112,122 +112,134 @@ func FromBytesBatchProof(b []byte) (BatchProof, error) {
...
+	// tuple[0] is the parent, tuple[1] is the left child and tuple[2] is the right child.
+	// trees holds the entire proof tree of the batchproof in this way, sorted by the tuple[0].
+	trees := make([][3]node, 0, len(computablePositions))
+	// initialise the targetNodes for row 0.
+	// TODO: this would be more straight forward if bp.Proofs wouldn't contain the targets
+	proofHashes := make([]Hash, 0, len(proofPositions))
+	targets := bp.Targets
+	var targetsMatched uint64
+	for len(targets) > 0 {
+		if targets[0] == numLeaves-1 && numLeaves&1 == 1 {
```
kcalvinalvin, line 149: Could you also leave a comment about the ```if``` clause that this code is checking for a row 0 root and how it's checking for them?

accumulator/batchproof.go
```diff
@@ -112,122 +112,134 @@ func FromBytesBatchProof(b []byte) (BatchProof, error) {
...
+			// row 0 root.
+			rootCandidates = append(rootCandidates, node{Val: roots[0], Pos: targets[0]})
+			bp.Proof = bp.Proof[1:]
+			break
+		}
 
-	tagRow := bp.Targets
-	nextRow := []uint64{}
-	sortUint64s(tagRow) // probably don't need to sort
+		if uint64(len(proofPositions)) > targetsMatched &&
```
kcalvinalvin, line 156: Could you add a comment on what this block is achieving?

accumulator/batchproof.go
```diff
@@ -112,122 +112,134 @@ func FromBytesBatchProof(b []byte) (BatchProof, error) {
...
+			bp.Proof = bp.Proof[2:]
+			targets = targets[1:]
+			continue
+		}
 
-	// TODO it's ugly that I keep treating the 0-row as a special case,
-	// and has led to a number of bugs.  It *is* special in a way, in that
-	// the bottom row is the only thing you actually prove and add/delete,
-	// but it'd be nice if it could all be treated uniformly.
+		if len(bp.Proof) < 2 || len(targets) < 2 {
```
kcalvinalvin, line 167: What would be a case where this clause is true?

accumulator/batchproof.go
```diff
@@ -112,122 +112,134 @@ func FromBytesBatchProof(b []byte) (BatchProof, error) {
...
+		// no roots to verify
+		return false, nil, nil
+	}
+	rootMatches := 0
+	for _, root := range roots {
+		if len(rootCandidates) > rootMatches && root == rootCandidates[rootMatches].Val {
+			rootMatches++
+		}
+	}
+	if len(rootCandidates) != rootMatches {
```
kcalvinalvin, line 237: Since all roots should match in order for the batchProof to be valid, wouldn't it be better to return false in the block at line 232 if any ```root``` doesn't match a ```rootCandidate```?

accumulator/pollardproof.go
```diff
@@ -7,204 +7,95 @@ import (
...
-// it returns a bool of whether the proof worked, and a map of the sparse
-// forest in the blockproof
-func (p *Pollard) verifyBatchProof(
-	bp BatchProof, roots []Hash,
-	numLeaves uint64, rows uint8) (bool, map[uint64]Hash) {
-
-	// if nothing to prove, it worked
-	if len(bp.Targets) == 0 {
-		return true, nil
+func (p *Pollard) populate(root *polNode, pos uint64, trees [][3]node, polNodes []polNode) int {
```
kcalvinalvin, line 46: Could you make a comment here?

accumulator/batchproof.go
dergoegge: I think the goto is simpler here because a separate function would have to take a ton of arguments and return all these slices i am appending to. It's also a bit easier to read imo than a function call in this case because it just jumps 10 lines down.
I think this is one of those rare cases where a goto is ok but if you strongly disagree i can change it.

accumulator/batchproof.go
dergoegge: i added a comment there, let me know if that explains it.

accumulator/batchproof.go
dergoegge: `rootCandidates` only holds a subset of the roots, so it could happen that a root does not match a candidate but by making sure that all candidates were matched to some root it still works out.
also added a comment there.

accumulator/batchproof.go
adiabat: How about just an `else`?  That seems even easier to understand

```
			proofPositions = proofPositions[1:]
			bp.Proof = bp.Proof[1:]
			targetNodes = targetNodes[1:]
		} else {
         		// target should have its sibling in targetNodes
	        	if len(targetNodes) == 1 {
		        	// sibling not found
	        		return false, nil, nil
	        	}
        		proof = targetNodes[1]
	        	targetNodes = targetNodes[2:]
		}
		left := target
		right := proof
```

dergoegge: changed it to and `else`.

kcalvinalvin:
Is there any code that would cause backwards incompatibility with master? Right now, It's fine when I point the csn to my local bridgenode based off of this PR but crashes with this message when I point it at the current default server.

```
[N] calvin@bitcoin ~/b/u/g/s/g/m/u/c/utreexoclient> ./utreexoclient -checksig=0
Creating new pollarddata
Verify failed: txo 6749762ae220c10705556799dcec9bb6a54a7b881eb4b961323a3363b00db518:0 position 103 leafdata 7b9562ebac55529d473d4d0bb616ee5d9f69898b1cc72051c204321666abea6d proof b628e99d6a24cbb09b35253756439ba827d27bc0b20c0d02a7bed7fc1dae4287
sib exists, af6aa1d9405b73e596cb3251c357f7dfdbbbece7f2db02176440098482bd64a2
panic: height 381 LeafData / Proof mismatch

goroutine 9 [running]:
github.com/mit-dci/utreexo/csn.(*Csn).IBDThread(0xc00000c1e0, 0xc000034150)
        /home/calvin/bitcoin-projects/utreexo1/go/src/github.com/mit-dci/utreexo/csn/ibd.go:56 +0xa29
created by github.com/mit-dci/utreexo/csn.(*Csn).Start
        /home/calvin/bitcoin-projects/utreexo1/go/src/github.com/mit-dci/utreexo/csn/init.go:101 +0x243
```

dergoegge:
> Is there any code that would cause backwards incompatibility with master?

That's a great test. The roots should be exactly the same for every block.

I think it errors because on master right now the 0-tree root is being send.
https://github.com/mit-dci/utreexo/blob/7f5d091c439e3255451009a4102a809c3b1b2ae6/accumulator/forestproofs.go#L195-L201

In this PR however those are not send/expected by the csn. I will check if that is actually the problem here.

PR #189, opened 2020-08-21
pollard: Make addOne remember leaf siblings correctly

dergoegge:
There was a bug in `addOne` that made the pollard not remember a leaf, see #180 for details.
This PR also refactors the roots to be pointers to `polNode` and makes readPos not stop at nil siblings.


adiabat:
Tested (but not the caching) & looks good

PR #192, opened 2020-08-25
accumulator: Add copy-on-write forestdata

kcalvinalvin:
Current Forestdata implementations have several downfalls.

1. Speed
2. Memory requirements
3. Fragile disk-storage

CowForest is faster than diskforest and cacheforest and trades off a little bit of overhead for a lot of memory savings.

On top of that, it adds safety so that in a case of a os-level crash, cowForest is able to resume from the latest commited forest state.

# Current code change that were major

1. `utreexoserver` now has 2 new flags:
  a. cow - enable cowForest (default false)
  b. cowmaxcache - how many max treeTables to keep in memory (default 500)

# CowForest

Quick benchmarks

Below is just a simple benchmark but is enough to show that cowForest is faster than disk and cache by a lot.

```
testnet3 to block 600,000. offsetfile pre-calculated.

inram         = 808.52user  89.54 -system  7:12.55 -elapsed  207%CPU
cow-no_sanity = 944.13user  91.25 -system  8:41.75 -elapsed  198%CPU
cow           = 977.29user  92.14 -system  9:07.24 -elapsed  195%CPU
cache         = 1031.32user 268.76-system  12:15.05-elapsed  176%CPU
disk          = 1089.61user 328.96-system  13:36.92-elapsed  173%CPU

*no-sanity is cowForest with sanity checks turned off. I think it's worth having on for now.
```

This is because all read and writes are done in memory with the necessary treeTables fetched from memory. 

CowForest is essentially a in-memory forest with some overheads for calculating local positions for treeBlock and treeTables, along with occasional flush and load from disk.

### Overall flow

**For `read()`:**

1. Calulate the positions of the TrekeTable and TreeBlock

2. Check if they're cached
  a. If they are, continue
  b. If they aren't, call `load()`
  and cache them

3. Calculate the local position of the leaf for the given treeBlock

4. Fetch and return hash

**For `write()`:**

1. Calculate the positions of the TreeTable and TreeBlock

2. Check if they're cached
  a. If they are, continue
  b. If they aren't, call `load()` and cache them.

3. Calulate the local position of the leaf for the given treeBlock

4. Replace hash in-memory

**For `swaphash()`:**

1. call `read()` on the leaves to be swaped and fetch the hash to be swapped

2. call `write()` on the relevant positions with the fetched hashes

**For `swapHashRange()`:**

1. Call `read()` on all the leaves to be swaped and fetch the hash to be swapped

2. Call `write()` on all the relevant positions with the fetch hashes

*Could potentially be faster if swapHashRange has it's own `readRange()` and `writeRange()` functions*

**For `size()`:**

1. Return (2 << cow.ForestRows) -1

*Really not needed*

**For `resize()`:**

1. Calculate how many treeBlockRowCount is needed to represent the forestRows.

Ex: For forestRows 19 and treeBlock has 6 forestRows, treeTables with treeBlocks with forestRows 0, 1, 2 are needed.

2. Loop through the treeBlockRowCount to see if there needs to be a new treeTable to accomodate the new size

**For `close()`:**

1. Call `commit()` and save the forest onto disk.

**Internal cowForest methods:**

**For `load()`:**

1. Check if the count of cached treeTables will be greater than the maxCount after the load.
  a. If it will be, call `flush()`

2. calculate the location of the treeTable

3. copy the file to a table in memory and add to cachedTreeMap. The key value is `fileNum++`

**For `flush()`:**

1. call `commit()`

2. reset `cow.cachedTreeTables` and `cow.meta.staleFiles`

3. load new manifest

**For `commit()`:**

1. commit manifest
2. save treeTables to disk
3. delete all staleFiles from disk


accumulator/forestdata.go
```diff
@@ -1,23 +1,55 @@
...
+	// returns how many leaves the current forest can hold
 	size() uint64
+
+	// allocate more space to the forest. newSize should be in leaf count (bottom row of the forest)
+	// can't resize down
 	resize(newSize uint64) // make it have a new size (bigger)
+
+	// just needed for cowForest since you need the forestRows value to
+	// calculate the leave positions for treeBlock
+	setRow(row uint8)
```
dergoegge, line 50: could this also be done through `resize`?

accumulator/forestdata.go
```diff
@@ -86,7 +118,1054 @@ func (r *ramForestData) close() {
...
+	}
+
+	return cow.manifest.fileNum
+}
+
+// Load will load the existing forest from the disk given a fileNumber
+func (cow *cowForest) load(fileNum uint64) error {
+	// check if there needs to be a flush
+	if len(cow.cachedTreeTables)+1 > cow.meta.maxCachedTreeTables {
+		cow.flush()
```
dergoegge, line 1020: is there maybe a better strategy than flushing everything?

accumulator/forestdata.go
```diff
@@ -86,7 +118,1054 @@ func (r *ramForestData) close() {
...
+
+// treeBlock is a representation of a row 6 utreexo tree.
+// It contains 127 leaves. The leaves can be a UTXO hash or another treeBlock
+type treeBlock struct {
+	// leaves are the utreexo nodes that are either
+	// 1: Hash of a UTXO
+	// 2: Parent hash of two nodes
+	//
+	// Able to hold up to 64 leaves
+	leaves [127]Hash
```
dergoegge, line 433: have you tried out different block sizes?
might be cool to have a benchmark comparing different sizes.

accumulator/forestdata.go
kcalvinalvin: Perhaps there is a better way but given 2 gigs for testnet3, it only flushed 3 times during the entire sync so I don't think it's the current bottleneck.

I wanted to get an initial version into master first and then work on improvements. So I think it's not bad putting this on the todo list for now.

accumulator/forestdata.go
kcalvinalvin: I have not yet. That's also something to try... I should change that 127 to a const variable though

accumulator/forestdata.go
kcalvinalvin: Fixed to use resize

PR #193, opened 2020-08-25
accumulator: Use sha512/256 instead of sha256

dergoegge:
I tested this up until block #300000 and sha512 seems to be a bit faster. (1.3x faster in my test)
[pprof_files.zip](https://github.com/mit-dci/utreexo/files/5126671/pprof_files.zip)


adiabat:
Cool that's an easy speedup :)

PR #196, opened 2020-09-03
Split proof building and serving functions

adiabat:
This PR should help with faster testing as the server binary can build a small number of proofs, quit, and then start serving them.  Generating the proofs and serving them are pretty independent things and this helps pull them apart.

### DIFF
--- a/bridgenode/server.go
+++ b/bridgenode/server.go
@@ -80,7 +80,6 @@ func stopServer(sig, haltRequest, haltAccept chan bool) {
 // ublocks blocks over that connection
 func blockServer(endHeight int32, dataDir string, haltRequest,
 	haltAccept chan bool, lvdb *leveldb.DB) {
-	endHeight-- // TODO make sure the -1 makes sense here
 	fmt.Printf("serving up to & including block height %d\n", endHeight)
 	listenAdr, err := net.ResolveTCPAddr("tcp", "0.0.0.0:8338")
 	if err != nil {
--- a/bridgenode/server.go
+++ b/bridgenode/server.go
@@ -14,7 +14,7 @@ import (
 	"github.com/syndtr/goleveldb/leveldb/opt"
 )
 
-func ServeBlock(param chaincfg.Params, dataDir string, sig chan bool) error {
+func ArchiveServer(param chaincfg.Params, dataDir string, sig chan bool) error {
 
 	// Channel to alert the tell the main loop it's ok to exit
 	haltRequest := make(chan bool, 1)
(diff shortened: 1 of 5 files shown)


cmd/utreexoserver/bridgeserver.go
```diff
@@ -114,13 +117,22 @@ func main() {
...
+		err := bridge.BuildProofs(param, dataDir, *forestInRam, *forestCache, sig)
+		if err != nil {
+			fmt.Printf("Buildproofs error: %s\n", err.Error())
+			panic("proof build halting")
+		}
+	}
+	err := bridge.ArchiveServer(param, dataDir, sig)
 	if err != nil {
-		fmt.Printf("Buildproofs error: %s\n", err.Error())
+		fmt.Printf("ServeBlocks error: %s\n", err.Error())
```
ysangkok, line 133: where does the name "ServeBlocks" come from? wouldn't it be better to call it "ArchiveServer error" since that is the function that errored?

adiabat: Ah yup, missed that! thanks

bridgenode/server.go
```diff
@@ -5,15 +5,83 @@ import (
...
+	<-haltAccept
+	os.Exit(0)
+}
+
 // blockServer listens on a TCP port for incoming connections, then gives
 // ublocks blocks over that connection
 func blockServer(endHeight int32, dataDir string, haltRequest,
 	haltAccept chan bool, lvdb *leveldb.DB) {
+	endHeight--
+	fmt.Printf("serving up to block height %d\n", endHeight)
```
ysangkok, line 84: maybe it would be helpful to say "up to and including", just to be explicit

bridgenode/server.go
```diff
@@ -5,15 +5,83 @@ import (
...
+
+	_, err := os.Stat(dataDir)
+	if os.IsNotExist(err) {
+		return fmt.Errorf("%s not found, can't serve blocks\n", dataDir)
+	}
+
+	// TODO ****** server shouldn't need levelDB access, fix this
+	ttlpath := "utree/" + param.Name + "ttldb"
+	// Open leveldb
+	o := opt.Options{CompactionTableSizeMultiplier: 8}
```
ysangkok, line 36: if the value 8 is arbitrarily chosen, maybe it would make sense to point that out

adiabat: True... but we need to get rid of levelDB there next; better to get rid of the whole thing!

bridgenode/server.go
adiabat: OK changed it... and the line before

adiabat:
Also now the auto testing takes 15 seconds instead of 3 minutes so that's nice, get the green check (hopefully) faster

PR #198, opened 2020-09-15
pollard: Add pruning to deletion

dergoegge:


accumulator/pollard.go
```diff
@@ -239,6 +234,7 @@ func (p *Pollard) rem2(dels []uint64) error {
 			// fmt.Printf("giving hasher %d %x %x\n",
 			// hn.position, hn.sib.niece[0].data[:4], hn.sib.niece[1].data[:4])
 			hn.dest.data = hn.sib.auntOp()
+			hn.sib.prune()
```
dergoegge, line 237: this does not quite work since it might prune a node that should become a new root.
is there a way to tell if the hn or it's children are future roots? maybe but swaps are still happening so hard to tell.

dergoegge: could also try to traverse the pollard after new roots are selected and prune but it would be cooler to not do that.

dergoegge: i think i figured it out. there were hashable nodes above the new roots, but those are not needed. excluding those nodes makes pruning in this place possible.

accumulator/pollard.go
```diff
@@ -225,6 +225,7 @@ func (p *Pollard) rem2(dels []uint64) error {
 		}
 		hashDirt = nextHashDirt
 		nextHashDirt = []uint64{}
+		// fmt.Printf("done with row %d %s\n", h, p.toString())
```
kcalvinalvin, line 228: I think this was removed in a previous commit. Could you remove this from your branch

kcalvinalvin:
LGTM. Maybe the whole proof should be remembered... Something to think about

dergoegge:
https://github.com/mit-dci/utreexo/blob/fbd19fb932017940cc8aa56a74b5d830368768ca/accumulator/pollard.go#L230-L238

This now happens for roughly 20% of the hashable nodes which (i think) means 20% less hashing but it would be cool if these duplicates would not be added to `hnslice` in the first place.
This happens so often now because the first time a hn for a pos shows up it might get pruned which makes it invalid the second time it shows up. There was no pruning before this PR. There is probably still duplicated hashing going on for nodes that are part of a remembered proof.

PR #206, opened 2020-10-01
Move TTL data to flat file

adiabat:
This gets rid of levelDB for the server by putting the TTL data into the flat file with the proofs.

### DIFF
--- a/util/types.go
+++ b/util/types.go
@@ -24,29 +24,26 @@ type ProofAndHeight struct {
 	Height int32
 }
 
-// Tx defines a bitcoin transaction that provides easier and more efficient
-// manipulation of raw transactions.  It also memoizes the hash for the
-// transaction on its first access so subsequent accesses don't have to repeat
-// the relatively expensive hashing operations.
-type zProofTx struct {
-	msgTx         *wire.MsgTx // Underlying MsgTx
-	txHash        *Hash       // Cached transaction hash
-	txHashWitness *Hash       // Cached transaction witness hash
-	txHasWitness  *bool       // If the transaction has witness data
-	txIndex       int         // Position within a block or TxIndexUnknown
-}
-
 // UBlock is a regular block, with Udata stuck on
 type UBlock struct {
-	Block     wire.MsgBlock
-	ExtraData UData
-	Height    int32
+	UtreexoData UData
+	Block       wire.MsgBlock
 }
 
+/*
+Ublock serialization
+(changed with flatttl branch)
+
+A "Ublock" is a regular bitcoin block, along with Utreexo-specific data.
+The udata comes first, and the height and leafTTLs come first.
+
+*/
+
 type UData struct {
-	AccProof accumulator.BatchProof
+	Height   int32
+	LeafTTLs []int32
 	UtxoData []LeafData
-	LeafTTLs []uint32
+	AccProof accumulator.BatchProof
 }
 
 // LeafData is all the data that goes into a leaf in the utreexo accumulator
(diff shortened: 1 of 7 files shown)


bridgenode/flatfileworker.go
```diff
@@ -0,0 +1,200 @@
...
+				maxOffsetHeight++
+			}
+			// for all the TTLs, seek and overwrite the empty values there
+			for _, c := range ttlRes.Created {
+				// seek to the location of that txo's ttl value in the proof file
+
+				// fmt.Printf("want ram offset %d, only have up to %d\n",
+				// 	c.createHeight, len(inRamOffsets))
+
+				_, _ = proofFile.Seek(
```
dergoegge, line 188: i would not be surprised if this slows things down quite a bit...
keeping the proofs&ttls for the last n blocks in memory could help but lets see if this is actually a problem first.

accumulator/batchproof.go
```diff
@@ -14,99 +14,126 @@ type BatchProof struct {
...
-	// if empty slice, return empty BatchProof with 0 targets
-	if len(b) == 0 {
-		return bp, nil
+// Deserialize gives a block proof back from the serialized bytes
+func (bp *BatchProof) Deserialize(r io.Reader) (err error) {
+	var numTargets, numHashes uint32
+	err = binary.Read(r, binary.BigEndian, &numTargets)
+	if err != nil {
+		if err == io.EOF && numTargets == 0 {
+			err = nil // EOF at the end is not an error...
```
ysangkok, line 78: why is numTargets being compared to zero?  the [docs](https://golang.org/pkg/encoding/binary/#Read) say that it returns EOF only if nothing at all was read. so i think this extra condition should only be necessary to prevent partial reading?

ysangkok (changes requested):

accumulator/batchproof.go
adiabat: Yeah that shouldn't be there, taken out.  With reader.Read that can return EOFs with a successful read but not binary.Read.

kcalvinalvin:
https://github.com/kcalvinalvin/utreexo/tree/calvin-flatttl

I've squashed the commits and fixed the unit tests in accumulator. `ud.Deserialize()` in `bridgenode/server.go` is broken though.

util/types.go
```diff
@@ -59,264 +56,242 @@ type LeafData struct {
...
-
-	udataBytes := make([]byte, uDataLen)
+	// fmt.Printf("read ttls %d\n", numTTLs)
+	// fmt.Printf("UData deser read h %d - %d ttls ", ud.Height, numTTLs)
 
-	for bytesRead < uDataLen {
-		n, err = r.Read(udataBytes[bytesRead:])
+	ud.TxoTTLs = make([]int32, numTTLs)
+	for i, _ := range ud.TxoTTLs { // write all ttls
+		err = binary.Read(r, binary.BigEndian, ud.TxoTTLs[i])
```
dergoegge, line 233: the ttl has to be passed as a pointer:
```suggestion
		err = binary.Read(r, binary.BigEndian, &ud.TxoTTLs[i])
```

it would also be possible to pass the slice directly:
https://github.com/golang/go/blob/d9725f549f48ae6e4eaf70311f6827173453935b/src/encoding/binary/binary.go#L210

dergoegge (changes requested):

util/types.go
adiabat: Oh man missed that.  And somehow that only causes it to fail intermittently / rarely?  Surprising

PR #208, opened 2020-10-06
test: Add csn resume test

### LINKED ISSUE #199: Add stop / resume to integration tests
Right now the integration tests does 1 pass without stopping.
We should test breaking the IBDsim (ctrl break after a timeout?) and resume it, maybe a few times, to test resuming.

dergoegge:
Closes #199

### DIFF
--- a/cmd/utreexoclient/csnclient.go
+++ b/cmd/utreexoclient/csnclient.go
@@ -50,6 +50,8 @@ var checkSig = optionCmd.Bool("checksig", true,
 	`check signatures (slower)`)
 var lookahead = optionCmd.Int("lookahead", 1000,
 	`size of the look-ahead cache in blocks`)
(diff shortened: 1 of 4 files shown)


kcalvinalvin:
I was thinking maybe we could do the quit with the linux `timeout` command. This way, we could do something like ```eval timeout -s INT 60 $IBDSIM --host=127.0.0.1```. This way, we don't touch the actual go stuff and I think it can be a simpler implementation overall.

kcalvinalvin (commented):
Not part of the code here but can we change this line https://github.com/mit-dci/utreexo/blob/92d6ee46da74f1304dd15a62c57c3ceb221b357c/test/csn_bridge.sh#L7-L8 to say `./utreexoserver` and `./utreexoclient` while we're changing this file

Issue #181, opened 2020-08-05
No restart when switching networks


fanquake:
When switching networks the client currently just resumes with the data of the previous network. i.e testnet -> mainnet:
```bash
./utreexoclient 
Creating new pollarddata
got to height 1000
got to height 2000
got to height 3000
got to height 4000
got to height 5000
got to height 6000
got to height 7000
got to height 8000
got to height 9000
Block 10000 add 24568 del 17548 pol nl 7020 roots 8 he 24546 re 0 ow 17666 
 plus 0.00 total 13.75 
got to height 10000
got to height 11000
got to height 12000
got to height 13000
got to height 14000
got to height 15000
got to height 16000
got to height 17000
got to height 18000
got to height 19000
Block 20000 add 35044 del 19170 pol nl 15874 roots 6 he 35021 re 0 ow 21964 
 plus 0.00 total 14.53 
got to height 20000
got to height 21000
got to height 22000
got to height 23000
got to height 24000
got to height 25000
got to height 26000
^CUser exit signal received. Exiting...
Block 26634 add 42780 del 20446 pol nl 22334 roots 10 he 42726 re 0 ow 24974 
 plus 0.00 total 15.23 
Pollard leaves: 22334
Found 0 satoshis in 0 utxos
Done Writing
....
./utreexoclient -net=mainnet
Has access to forestdata, resuming
Restoring Pollard Roots...
22334 leaves 10 roots Finished restoring pollard
got to height 27000
got to height 28000
got to height 29000
Block 30000 add 4454 del 665 pol nl 26123 roots 7 he 4441 re 0 ow 1290 
 plus 0.00 total 0.33 
got to height 30000
got to height 31000
got to height 32000
got to height 33000
got to height 34000
got to height 35000
got to height 36000
got to height 37000
got to height 38000
got to height 39000
Block 40000 add 31266 del 13030 pol nl 40570 roots 10 he 31299 re 0 ow 19589 
 plus 0.00 total 4.86 
got to height 40000
^CUser exit signal received. Exiting...
Block 40394 add 32278 del 13589 pol nl 41023 roots 8 he 32321 re 0 ow 20350 
 plus 0.00 total 5.10 
Pollard leaves: 41023
Found 0 satoshis in 0 utxos
Done Writing

kcalvinalvin:
Hrm yeah we should save network metadata somewhere

adiabat:
[...]

PR #215, opened 2020-10-31
bridgenode: Make paths separate for different networks

kcalvinalvin:
Main changes:
1. All files in the `utree/` directory is now just saved to a default path of `$HOME/.utreexo`. Handles windows and osx as well. To specify a custom path, pass in flag `-bridgedir` to the command. `testnet` files are stored in `.utreexo/testnet3/` and `regtest` files are stored in `.utreexo/regtest`. `mainnet` files are just in `.utreexo` like how bitcoind does it.
2. Merged `cow`, `cache`, `inram` to be under one flag `forest`. User must now specify which forest they want. Ex: `forest=cow`. Typos throw an error.
3. A new `config` struct which has *most* the configs needed for the bridgenode. This is a start to no more parsing and configuration for individual functions.
4. Moves the arg-parse logic from `cmd/` to `bridgenode`

resolves issue #181 

EDIT: Added a test that compares proofs with bridgenode that doesn't stop and resume with one that does. When #219 is merged, the tests pass.


bridgenode/initrestore.go
```diff
@@ -28,43 +23,39 @@ func initBridgeNodeState(
...
 		if err != nil {
 			err = fmt.Errorf("createOffsetData error: %s", err.Error())
 			return
 		}
 		fmt.Printf("tip height %d\n", knownTipHeight)
 	}
 
-	if cowForest {
-		if util.HasAccess(util.CowForestCurFilePath) {
+	if cfg.cowForest {
```
dergoegge, line 43: The if/else blocks look very similar and could be cleaned up. They only differ in the file that gets passed to `util.HasAccess`. Maybe something for another PR, since this PR did not introduce this.

dergoegge (approved):
ACK https://github.com/mit-dci/utreexo/pull/215/commits/dd0fb0e8d945e16e6c4096cb5b3cf001c269fa99

This is nice because now all the data is stored separately from the bitcoin data dir. Also switching networks now works as intended!

bridgenode/initrestore.go
kcalvinalvin: Ah yeah not the best. I'll clean this up with this PR since it is sorta a cleanup PR anyways

kcalvinalvin: Fixed with a `forestType` enum

dergoegge:
reACK https://github.com/mit-dci/utreexo/commit/49a65349691e142570953f121ad0a69faafd817d

The compare test is a nice addition. Verified that it passes with the changes of #219.

PR #217, opened 2020-11-02
WIP: Network btcd

kcalvinalvin:
Seems like most of the peer/netsync/connmgr packages can be reused. Just need to implement the message handlers with the accumulator stuff


kcalvinalvin:
Implemented in github.com/mit-dci/utcd

PR #218, opened 2020-11-02
accumulator: Fix batch proof verification

dergoegge:
See commit message for explanation.

I checked with the previous map based verification and the same problem existed there:
https://github.com/mit-dci/utreexo/blob/573edd2bb6d34f7073873e617c13c056e745d658/accumulator/pollardproof.go#L177-L194

`parhash` is considered to be equal to the result of `parentHash(...)`.

The main problem here is that the roots are always cached and therefore the roots computed from the proof are never actually compared to the actual roots.

### DIFF
--- a/accumulator/pollard_test.go
+++ b/accumulator/pollard_test.go
@@ -33,6 +33,33 @@ func TestPollardFixed(t *testing.T) {
 	}
 }
 
+func TestPollardSimpleIngest(t *testing.T) {
+	f := NewForest(nil, false, "", 0)
+	adds := make([]Leaf, 15)
+	for i := 0; i < len(adds); i++ {
+		adds[i].Hash[0] = uint8(i + 1)
+	}
+
+	f.Modify(adds, []uint64{})
+	fmt.Println(f.ToString())
+
+	hashes := make([]Hash, len(adds))
+	for i := 0; i < len(hashes); i++ {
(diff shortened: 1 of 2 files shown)


accumulator/pollard_test.go
```diff
@@ -33,6 +33,33 @@ func TestPollardFixed(t *testing.T) {
...
+
+	bp, _ := f.ProveBatch(hashes)
+
+	var p Pollard
+	p.Modify(adds, nil)
+	// Modify the proof so that the verification should fail.
+	bp.Proof[0][0] = 0xFF
+	err := p.IngestBatchProof(bp)
+	if err == nil {
+		t.Fatal(err)
```
kcalvinalvin, line 59: If this test does fail, it only prints that the error is nil like so:
```
--- FAIL: TestPollardSimpleIngest (0.00s)
    pollard_test.go:59: <nil>
FAIL
exit status 1
```

Which is weird because a test is failing when error is nil. 

I think maybe something like `err = fmt.Errorf("BatchProof valid after modification. Accumulator validation failing")` would be better.

kcalvinalvin:
This is good, nice catch.

Verified that the same bug exists when map is used (I checked on commit `589529f56aa47a66566076abc66c3f66d0f8f2ab`)

PR #219, opened 2020-11-06
Fix bridge restart

### LINKED ISSUE #216: Consistently missing udata from server when stopping and resuming
On master, I get this message when syncing both the client.

```
panic: uData missing utxo data for block 47413

goroutine 66 [running]:
github.com/mit-dci/utreexo/csn.(*Csn).IBDThread(0xc00000c1e0, 0xc000524000, 0xffffffffffffffff)
        /home/calvin/bitcoin-projects/go/utreexo/go/src/github.com/mit-dci/utreexo/csn/ibd.go:59 +0xbe5
created by github.com/mit-dci/utreexo/csn.(*Csn).Start
        /home/calvin/bitcoin-projects/go/utreexo/go/src/github.com/mit-dci/utreexo/csn/init.go:102 +0x275
```

block 47413 was the block that was stopped and resumed from the server side.

dergoegge:
The offset of the last written proof was saved which caused an off-by-one on restart because that last proof is over written by the next block.

Closes #216

### DIFF
--- a/bridgenode/flatfileworker.go
+++ b/bridgenode/flatfileworker.go
@@ -158,6 +158,10 @@ func (ff *flatFileState) ffInit() error {
 			ff.offsets[ff.currentHeight] = ff.currentOffset
 			ff.currentHeight++
 		}
+
+		// set currentOffset to the end of the proof file
+		ff.currentOffset, _ = ff.proofFile.Seek(0, 2)
+
 	} else { // first time startup
 		// there is no block 0 so leave that empty
 		fmt.Printf("setting h=1\n")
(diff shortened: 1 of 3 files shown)


kcalvinalvin:
Hmm seems like the proofs still differ for when you stop and restart vs when you just let the bridgeserver complete by itself.

I've added a break at height 100,000 in the main loop at `genproofs.go` like so
```
               if height == 100000 {
                       break
               }
```

Then I diff `proof.dat` vs when I didn't stop and when I did stop in between. I get different files
```
> diff completedtree/proofdata/proof.dat  tmptree/proofdata/proof.dat
Binary files utree/proofdata/proof.dat and nostoptree/proofdata/proof.dat differ
```

kcalvinalvin:
ACK ea0e1195cf99a7c46fabfc5241d8a000f4f0ae2a

Tested and works. `proof.dat` files are the same.

PR #222, opened 2020-11-11
accumulator: Check number of proof hashes

dergoegge:
If the number of proof positions (+number of targets) does not match the number of provided hashes the proof is invalid.


kcalvinalvin:
ACK. Much better than just panicking.

Just to verify though, was there a case where the verify passed if the if branch wasn't present?

dergoegge:
@kcalvinalvin afaik no. In my testing it would just crash.

PR #224, opened 2020-11-12
Actions: Compare proofs with master

kcalvinalvin:
Fetches the master branch and diffs the proof against the one generated by the PR.

It's currently breaking because the master binary doesn't have the flags I'm using in the script. Works after #215 is merged.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -6,6 +6,12 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5
     steps:
+
+    - name: setup directory
+      shell: bash
+      run: |
+        mkdir -p "${{ github.workspace }}/testdatapath"
+
     # set GOPATH
     - name: setup env
       shell: bash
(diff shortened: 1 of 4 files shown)


test/csn_bridge.sh
```diff
@@ -10,7 +10,8 @@ set -Eeuo pipefail
 
 GENPROOFS="$1"
 IBDSIM="$2"
-TEST_DATA=$(mktemp -d)
+TEST_DATA="$3"
```
dergoegge, line 13: Could still assign a temp dir if $3 is not given.

.github/workflows/go.yml
```diff
@@ -54,4 +60,63 @@ jobs:
...
-        ./test/csn_bridge.sh "./cmd/utreexoserver/utreexoserver" "./cmd/utreexoclient/utreexoclient"
+        ./test/csn_bridge.sh "./cmd/utreexoserver/utreexoserver" "./cmd/utreexoclient/utreexoclient" "${{ github.workspace }}/testdatapath"
+
+    - name: Upload testdata
+      uses: actions/upload-artifact@v2
+      with:
+        name: testdatapath
+        path: "${{ github.workspace }}/testdatapath"
+
+  check-proof-backwards:
```
dergoegge, line 71: It looks a bit weird that there are so many duplicate steps, like "setup env", "Install Go", "Get deps" and "Build" are the same as in the previous job. 
Could this not just be a couple extra steps in the first job?
After "Run integration tests" add: "Checkout master", "Build", "Check backwards".

.github/workflows/go.yml
```diff
@@ -54,4 +60,63 @@ jobs:
...
+        cd $GOPATH/src/github.com/${{ github.repository }}
+        go get -v -d ./...
+
+    - name: Build
+      shell: bash
+      run: |
+        cd $GOPATH/src/github.com/${{ github.repository }}/cmd/utreexoserver
+        go install -v
+
+    - name: checkout
```
dergoegge, line 112: Is this only needed because `./test/check_proofs_backwards.sh` does not yet exist on master?

kcalvinalvin: Yes. Not exactly the prettiest... I guess we could update this afterwards

test/csn_bridge.sh
kcalvinalvin: Ok would be helpful for testing locally. I'll add this

.github/workflows/go.yml
kcalvinalvin: Jobs don't share state so they would have to be uploaded to github artifacts.

I think it'd be more maintainable/functional if the second job fetches and builds itself. That process is fairly quick as well

test/csn_bridge.sh
kcalvinalvin: Added

PR #225, opened 2020-11-12
accumulator/hasher.go: Faster hashrow

kcalvinalvin:
EDIT: sha512 was not the culprit, it was the goroutines.

`hashOne()` was taking up 22% of the overall sync time from my tests up to height `150,000` on mainnet.
Screen Shot 2020-11-12 at 5 33 41 PM

This slowdown is mostly caused by the heap allocations that each of the goroutine makes. Each makes 4096bytes of stack so over time this builds up a lot. Solution was to just get rid of the goroutines.

Profiling for the new code looks like so. `root` takes 142.15 seconds. Note how long the `hashRow()` function takes. This is with the individual hashes counting towards the overall time (since there is no more goroutines).
Screen Shot 2020-11-12 at 5 34 41 PM

Compared to the old profile here, where the `root` takes 250.44 seconds. Note how the `hashRow()` function takes longer, not counting the actual individual hashes.
Screen Shot 2020-11-12 at 5 34 30 PM

### DIFF
--- a/accumulator/hasher.go
+++ b/accumulator/hasher.go
@@ -6,32 +6,11 @@ type hashableNode struct {
 	position  uint64 // doesn't really need to be there, but convenient for debugging
 }
 
-type hashNpos struct {
-	result Hash
-	pos    uint64
-}
-
-func hashOne(l, r Hash, p uint64, hchan chan hashNpos) {
-	var hnp hashNpos
-	hnp.pos = p
-	hnp.result = parentHash(l, r)
-	hchan <- hnp
-}
-
 func (f *Forest) hashRow(dirtpositions []uint64) error {
-
-	hchan := make(chan hashNpos, 256) // probably don't need that big a buffer
-
 	for _, hp := range dirtpositions {
 		l := f.data.read(child(hp, f.rows))
 		r := f.data.read(child(hp, f.rows) | 1)
-		// fmt.Printf("hash pos %d l %x r %x\n", hp, l[:4], r[:4])
-		go hashOne(l, r, hp, hchan)
-	}
-
-	for remaining := len(dirtpositions); remaining > 0; remaining-- {
-		hnp := <-hchan
-		f.data.write(hnp.pos, hnp.result)
+		f.data.write(hp, parentHash(l, r))
 	}
 
 	return nil


kcalvinalvin:
Also there's probably a lot we can gain on pollard by doing this...

adiabat:
But wait, with this PR it just does every hash sequentially, right?

I can see if the `go hashOne()` calls are slowing things down, but there must be another way to do multi-threaded hashing that's faster.  I guess we can merge this as it's basically 3 lines and easy to change... but what about other ways to do it?

If creating & destroying goroutines is too costly, what about instead of 1 goroutine per hash, we have a number of worker routines (eg same as numcpus) that are all pulling from the same channel.

Since you have all the testing set up, how about I paste the code here and compare that as well?

adiabat:
OK instead of PRing your PR, I'll just paste the whole file here:

```
package accumulator

// hashableNode is the data needed to perform a hash
type hashableNode struct {
	sib, dest *polNode
	position  uint64 // doesn't really need to be there, but convenient for debugging
}

type hashWork struct {
	l, r Hash
	pos  uint64
}

func hasherWorker(workChan chan hashWork, resultChan chan hashWork) {
	for {
		w, stillOpen := <-workChan
		if !stillOpen {
			return
		}
		w.r = parentHash(w.l, w.r)
		resultChan <- w
	}
}

func (f *Forest) hashRow(dirtpositions []uint64) error {

	hworkChan := make(chan hashWork, 256) // probably don't need that big a buffer
	hresultChan := make(chan hashWork, len(dirtpositions))

	for numThreads := 0; numThreads < 4; numThreads++ {
		go hasherWorker(hworkChan, hresultChan)
	}

	for _, hp := range dirtpositions {
		var hw hashWork
		hw.pos = hp
		hw.l = f.data.read(child(hp, f.rows))
		hw.r = f.data.read(child(hp, f.rows) | 1)
		hworkChan <- hw
	}
	close(hworkChan)

	for remaining := len(dirtpositions); remaining > 0; remaining-- {
		hnp := <-hresultChan
		f.data.write(hnp.pos, hnp.r)
	}

	return nil
}
```

can you compare this to the too many goroutines and the single threaded versions and see how it does?

(Also I'm not 100% sure this works properly as I can't find the specification of the channel behavior here.  I'm closing it on line 41, and then checking if it's closed with `w, stillOpen := <-workChan`.  Can it still pull from the channel even when it's closed?  Does `stillOpen` return true after I've "closed" it but the channel still has stuff in the buffer?  It seems like it does because this seems to work, and if closing the channel killed everything in the buffer, or caused stillOpen to immediately return false, then it would skip some hashes and the proofs wouldn't work.  

And it seems like they do, so the `item, open := <- channel` syntax returns false only when the channel is closed and the buffer is empty.  But I couldn't find anything explicitly saying that that's how it works.  But pretty sure it does.

Anyway this seems like it would be faster, lower overhead, and not too complex, right?  If the code is efficient, hashing should be the bottleneck, and should be very parallelizable

kcalvinalvin:
Ok I benchmarked both the hashRows, one with the code suggested above and one with just sequential hashing.

Well the fun bit is that we know it's not the goroutines but the channel sends and receives. The sends and receives are expensive relative to the cost of a single hash.

Testing method:
1. mainnet
2. Up to block #150,000
3. Offsetdata pre-generated

## Wall clock speeds:
```
  Suggested code with one goroutine per row                                                                                                                                           
                                                                                                                                                                                
  real    1m14.735s                                                                                                                                                             
  user    3m28.218s                                                                                                                                                             
  sys     0m50.568s                                                                                                                                                             
                                                                                                                                                                                
  real    1m14.463s                                                                                                                                                             
  user    3m26.744s                                                                                                                                                             
  sys     0m50.391s                                                                                                                                                             
                                                                                                                                                                                
  real    1m14.708s                                                                                                                                                             
  user    3m27.656s                                                                                                                                                             
  sys     0m50.553s
```


```
  Code in this PR

  real    1m0.298s                                                                                                                                                              
  user    1m48.459s                                                                                                                                                             
  sys     0m42.366s                                                                                                                                                             
                                                                                                                                                                                
  real    1m0.777s                                                                                                                                                              
  user    1m48.667s                                                                                                                                                             
  sys     0m42.559s                                                                                                                                                             
                                                                                                                                                                                
  real    1m0.237s                                                                                                                                                              
  user    1m47.646s                                                                                                                                                             
  sys     0m42.392s
```

## Pprof flamegraphs

### Code suggested

Root speed:
<img width="1539" alt="Screen Shot 2020-11-13 at 5 10 00 PM" src="https://user-images.githubusercontent.com/37185887/99045163-44a6ab00-25d4-11eb-87fa-ca2a7175cfab.png">

hasherWorker
<img width="1533" alt="Screen Shot 2020-11-13 at 5 09 41 PM" src="https://user-images.githubusercontent.com/37185887/99045143-3d7f9d00-25d4-11eb-8115-572d6f165915.png">

Note the cost of channel sends/receives
<img width="1527" alt="Screen Shot 2020-11-13 at 5 09 52 PM" src="https://user-images.githubusercontent.com/37185887/99045298-73bd1c80-25d4-11eb-8840-b5ebe4771d4b.png">

## Code in this PR

Root speed:
<img width="1533" alt="Screen Shot 2020-11-13 at 5 10 07 PM" src="https://user-images.githubusercontent.com/37185887/99045327-833c6580-25d4-11eb-825f-0989b8b07696.png">

<img width="1547" alt="Screen Shot 2020-11-13 at 5 10 23 PM" src="https://user-images.githubusercontent.com/37185887/99045483-c1398980-25d4-11eb-8d2e-72aa718fac0a.png">

dergoegge:
The hasherWorker code could be optimised to only do as many channel send/recv as there are workers instead of one send/recv per hash, by passing a slice of hashWork to the workers.

adiabat:
OK merging for testing but we need to fix this to make it multithreaded

// TODO !!

PR #226, opened 2020-11-13
WIP: Pollard hashes up to populated nodes, not all the way to roots

adiabat:
Previously `(p *Pollard) IngestBatchProof()` would verify the batch proof, then apply it to the pollard; the verification step verified proofs all the way up to the roots.

This verification is inefficient as the pollard can already have hashes near the top of the tree.  Also if you need to hash up to roots, that means you can't have partial proofs and can't use the TTL values.

This PR will change pollard's `IngestBatchProof()` to only hash up to known values, and not require proofs up to the roots.

The simplest way is that it will *require* proofs up to exactly it's lowest populated position, which is fragile as it won't be able to accept a larger than needed proof.  But that seems like an OK first step, we can deal with excess proofs in a later PR.


dergoegge:
> The simplest way is that it will require proofs up to exactly it's lowest populated position, which is fragile as it won't be able to accept a larger than needed proof

This also has an upside because the csn can tell if the bridge sends to much. So it could disconnect if the bridge makes it download more than needed.

util/types.go
```diff
@@ -525,6 +526,7 @@ func (ub *UBlock) SerializeCompactSize() int {
 // rebuilt from the block data.  Note that this leaves the blockhash
```
dergoegge, line 542: The blockhash is not yet being restored right? at least i can't find it in any of the commits.
This also assumes a height to blockhash index on the client side exists, might not be worth it if the saving aren't huge.

PR #229, opened 2020-11-18
WIP: Faster row hashing during delete

dergoegge:
Benchmarks indicate that this is worth it:
```
$ go test -bench=RowHasher. -run=Bench
goos: darwin
goarch: amd64
pkg: github.com/mit-dci/utreexo/accumulator
BenchmarkRowHasherSequential-4           1876930               600 ns/op
BenchmarkRowHasherParallel-4             3982798               335 ns/op
PASS
ok      github.com/mit-dci/utreexo/accumulator  4.349s
```

### DIFF
--- a/accumulator/forest.go
+++ b/accumulator/forest.go
@@ -4,7 +4,9 @@ import (
 	"encoding/binary"
 	"fmt"
 	"os"
+	"runtime"
 	"sort"
+	"sync"
 	"time"
 )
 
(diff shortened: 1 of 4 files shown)


dergoegge:
After some investigation i found that most of the time its not worth it to parallelize the hashing as there are not enough hashes to be done on a row. On testnet for example the avg. number of hashes per worker with 4 workers is 0.2 (up to block 240k). I could imagine though that this is not the same on mainnet when blocks are full. 
`hashRow` now only splits up the hashing if there is atleast a certain number of hashes available for each worker.

Will do some profiling...

PR #230, opened 2020-11-20
Refactor csn and utils for exports

### LINKED ISSUE #195: Improvements: Clean up packages for easy imports to other codebases
Since the goal of the Utreexo project is to be adopted by other projects a good package with detailed documentation on the usage is a must.

## Support easy imports by other codebases

**EDIT 1: Changed the exported functions**

- [ ] **Decide on which methods/functions to be exported**

    Currently, there are many exported functions that aren't really relevant to the end user of Utreexo. There should only be a few exported functions that are well documented.

    * `OpenPollard(fileName string) (Pollard, error)` // Opens the pollard in the given path. Creates if one doesn't exist
    * `type Pollard struct`      
      * `Modify(dels []uint64, adds []Leaf) error`
      * `Verify(proof Proof) bool`
      * `Roots() []Hash`
      * `Flush(rootsOnly bool) error` flush the pollard to `file`, it's possible to flush the cached nodes and roots or just the roots

    * `OpenForest(fileName string) (Forest, error)` // Opens the forest in the given path. Creates if one doesn't exist
    * `type Forest struct`      
      * `Modify(dels []uint64, adds []Leaf) (UndoBlock, error)`
      * `CreateProof(targets []uint64) Proof`
    
    * `type Proof struct`
        * `FetchTargets() []uint64`

--
    To keep the accumulator library independent from Bitcoin, the accumulator functions should be wrapped in ```bridgenode``` and ```csn``` and these functions should be exported from there.

Google has the pkg.go.dev up and running now for a few months. godoc links directly to here and is used quite frequently by go developers. Our goal should be to make this easy readable and be importable by an external project.
https://pkg.go.dev/github.com/mit-dci/utreexo/accumulator?tab=doc

Looking for reviews for which functions to be exported. @adiabat @ysangkok @dergoegge 

- [ ] **Documentation improvements for pkg.go.dev**

    This goes in hand with the TODO above. When making a nice, clean functions ready to be imported, the documentation should be updated to match the code.

kcalvinalvin:
1. CSN parse logic switched to the bridgenode model where most of the required logic for parsing and starting a server is in the `csn` package.

2. New packages `wire` and `btcacc`. All the udata and leafdata are now moved to `btcacc` and shared functions for csn/ and bridgenode/ needed for wire is moving to the `wire` package.

Closes #195

A side note should be made that package `btcacc` cannot contain any packages from btcsuite because that will result in an import cycle in btcd when implementing Utreexo there. This is why some core functions/types (UBlock being a notable example) are not in btcacc and are written in the btcsuite packages themselves.


adiabat:
Looks good, tested.  Also yeah why was the actual pollard struct in pollardutil (shrug).

PR #232, opened 2020-12-01
Update go.mod

kcalvinalvin:
Fixes the build failure on master, which couldn't find the commit it was pointed to since I had squashed and pushed to the rjected repo.

Now that the one function we needed in btcd is merged, we can point to the btcd repo. Rjected repo is gonna be modified a lot as well so better to point directly to the btcd repo


adiabat:
OK for now... but we'll probably end up pointing back to rjected again right?
Or can make "utcd" or something weird.

kcalvinalvin:
Maybe? But yeah if we do need something for our use, we can always just add that one replace line back to the go.mod

PR #250, opened 2021-02-15
accumulator: Change getRootsReverse to getRootsForwards and add a freelist for positionlist

kcalvinalvin:
Changes made:

## getRootsForwards

getRootsReverse forces unnecessary allocations as the roots are
appended to a new list. This helps reduce the amount of memory
allocations, resulting in less time taken up by the gc.

Also, grabbing roots forwards is a lot cleaner than the reverse method.

## positionList

Positionlist (aka slice of uint64) is used numerous times in the accumulator code. Having a freelist for this list reduces gc pressure.

## computedPositions as int64 instead of a slice

All the functions that use computedPositions only take the length of the computedPositions slice. Changing this to just a int64 reduces gc pressure and speeds up the code.

### DIFF
--- a/accumulator/utils.go
+++ b/accumulator/utils.go
@@ -4,28 +4,60 @@ import (
 	"fmt"
 	"math/bits"
 	"sort"
+	"sync"
 )
 
 // verbose is a global const to get lots of printfs for debugging
 var verbose = false
 
+// PositionList is a wrapper around slice of ints.
+type PositionList struct {
+	list []uint64
+}
+
+// Free returns the slice of uint64 back to the pool
+func (pl *PositionList) Free() {
+	pl.list = pl.list[:0]
+	positionListFree.Put(pl)
+}
+
+var positionListFree = sync.Pool{
+	New: func() interface{} { return new(PositionList) },
+}
+
+func NewPositionList() *PositionList {
+	p := positionListFree.Get().(*PositionList)
+
+	if p.list == nil {
+		// allocate a little bit since 0 len is gonna be costly
+		// when a new PositionList is created
+		p.list = make([]uint64, 0, 30)
+	}
+
+	return p
+}
+
 // ProofPositions returns the positions that are needed to prove that the targets exist.
 func ProofPositions(
-	targets []uint64, numLeaves uint64, forestRows uint8) ([]uint64, []uint64) {
+	targets []uint64, numLeaves uint64, forestRows uint8, proofPositions *[]uint64) int64 {
 	// the proofPositions needed without caching.
-	proofPositions := make([]uint64, 0, len(targets)*int(forestRows))
 	// the positions that are computed/not included in the proof.
 	// (also includes the targets)
-	computedPositions := make([]uint64, 0, len(targets)*int(forestRows))
+
+	nextTargets := NewPositionList()
+	defer nextTargets.Free()
+
+	var computedPositions int64
 	for row := uint8(0); row < forestRows; row++ {
-		computedPositions = append(computedPositions, targets...)
+		computedPositions += int64(len(targets))
 		if numLeaves&(1<<row) > 0 && len(targets) > 0 &&
 			targets[len(targets)-1] == rootPosition(numLeaves, row, forestRows) {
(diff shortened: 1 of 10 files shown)


accumulator/utils.go
```diff
@@ -4,28 +4,60 @@ import (
...
 	"math/bits"
 	"sort"
+	"sync"
 )
 
 // verbose is a global const to get lots of printfs for debugging
 var verbose = false
 
+// PositionList is a wrapper around slice of ints.
+type PositionList struct {
```
dergoegge, line 14: nit: Could this not just be:
```go
type PositionList []uint64
```
to avoid code like: `positionList.list`

kcalvinalvin: I'll modify it and push it along with better getRootsReverse (and make getRootsForwards a separate function) as talked about in the utreexo meeting call.

kcalvinalvin: Ok so change of mind on the `type PositionList []uint64` because then the function would have to ask for `*PositionList` and not `*[]uint64`.

You could get around that problem by type casting but then you have to do this shimmy
```
rows := treeRows(numLeaves)
positionList := NewPositionList()
defer positionList.Free()

x := []uint64(*positionList)
ProofPositions(targets, numLeaves, rows, &x)
```

which arguably is worse than positionList.list

kcalvinalvin:
@adiabat I've done some more pondering and I think `getRootsForwards()` is a cleaner & better design vs `getRootsReverse()`.

So if we go back to why `getRootsReverse()` is slow, it's because of these two lines
```
roots = append([]uint64{root}, roots...)
rows = append([]uint8{row}, rows...)
```

Here the slice being appended to `[]uint64{root}` would be extended by `len(roots)` and the elements of roots would be appended to  `[]uint64{root}`. The element `roots` being used in `roots...` would then be collected by the gc. Since `getRootsReverse()` is called very often (once per loop in add), this is a massive overhead.

As you can see below, `getRootsReverse()` allocates the most memory out of all the functions in the utreexoserver code.

Screen Shot 2021-02-16 at 5 33 32 PM

This one function is responsible for ~12% of all the memory allocations for `utreexoserver` (The percentage is actually worse if you get rid of ttldb).

Screen Shot 2021-02-16 at 5 30 56 PM

Looking at the cpu profiling, we can see that `getRootsReverse()` is taking the majority of the time during `addv2()`, even more so than the hashing.

Screen Shot 2021-02-16 at 5 35 05 PM

To fix this, I can think of two solutions.
1: Have the slice pre-allocated and use indexes
2: Change the loop in `getRootsReverse()` to read from right to left, and not left to right.

2 is tricky since it's essentially re-implementing an existing algorithm that's fairly well tested now and wouldn't be the best use of time.

1, arguably in my opinion, isn't as good of a solution because of it's hard to allocate the slice from the caller side.

Take for example this loop
```
  // Add adds leaves to the forest.  This is the easy part.
  func (f *Forest) addv2(adds []Leaf) {
          // allocate the positionList first
          positionList := NewPositionList()
          defer positionList.Free()

          for _, add := range adds {
                  // reset positionList
                  positionList.list = positionList.list[:0]

                  f.positionMap[add.Mini()] = f.numLeaves
                  getRootsForwards(f.numLeaves, f.rows, &positionList.list)
...
```

This loop is efficient as the positionList is only borrowed once. And that positionList same is used for all the adds.

If we were to have `getRootsReverse()` allocate the slices inside the loop, then there are multiple allocations for a single call to `addv2()` vs the single `NewPositionList()` and reusing that.

If we were to pre-allocate from the caller-side, then there would have to be a slice-length check on the caller's side and another inside the `getRootsReverse()` function. I don't think this really worth the tradeoff.

The `getRootsForwards()` commit `28ac8ca296786e59de1b62ec6c9865c3b39b0aee` is fairly simple in that it changes some tests and some functions now using `roots[len(roots)-(i+1)]` instead of `roots[i]`. While this is more complicated, it's offset by the performance gain had by this.

I can make the functions separate but I really do think that `getRootsReverse()` should be the function that isn't called in this case.

Issue #261, opened 2021-04-04
Move all features in the utcd repo to the utreexo repo


kcalvinalvin:
There's a lot of btcd-ness in the utcd repo and the repo visibility isn't great. It's also harder to have utreexo node specific tests because of all the btcd-ness that we also have to keep in place.

One example: We need to have duplicate code for the wire/ stuff since go doesn't allow import cycles. This results in ugly mess in both sides of the repositories.

Moving all the utcd features that we lack here will help with test-ability, visibility, and better integration.

PR #262, opened 2021-04-05
wire: Modify and import the wire package from btcd

kcalvinalvin:
this PR mostly does two things

## 1. Import utcutil/ & wire/ into utreexo

Package `wire/` is taken from btcd and is trimmed of all the non-utreexo specific files (ex: compact filter related files). Then the varint serialization that was in the `wire/` package is removed and is placed in the current `util/` package that we have.

All the files that we need from `utcutil/` package is now placed into `chain/` package. This is now where the block, ublock, and tx lives.

## 2. Support Varint serialization

`UData`, `BatchProof`, and `Leaf` structs now all support `Encode`, `Decode`, and `SerializeSizeVarInt`. They differ from the current serialization/deserialization in that varint serialization is used for integers. 

This PR is one of the PRs to achieve #261


kcalvinalvin:
The CSN stuff is broken. I'll push a PR fixing this soon.

The wire/ and bridgenode/ packages won't be touched for the CSN fix.

kcalvinalvin:
Closing due to change in the direction of the development

PR #278, opened 2021-06-22
accumulator/batchproof: Remove target hashes from BatchProof

kcalvinalvin:
Before, the target hashes also used to be included in the BatchProof.
However, these are just the hash commitments of elements to prove and
the user has all the data needed to calculate the hashes.

Removing them achieves two things: smaller proofs, simpler verification.

Since the hashes no longer is being stored, the proofs get smaller by
however many deletions that are to happen. The verification logic is
simpler as it no longer has to have extra logic to separate out the
target hashes from the actual proofs.

### DIFF
--- a/accumulator/batchproof.go
+++ b/accumulator/batchproof.go
@@ -20,8 +20,9 @@ type BatchProof struct {
 	Targets []uint64
 
 	// All the nodes in the tree that are needed to hash up to the root of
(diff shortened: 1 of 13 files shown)


kcalvinalvin:
Ok urg the cmd/ stuff depends on utcd/blockchain and that's not building because this is a breaking change. On it...

PR #280, opened 2021-06-28
accumulator&bridgenode: Reset maps to avoid out of memory errors

kcalvinalvin:
# First change

Go maps don't shrink when elements are deleted from the map, resulting in maps growing as time passes. This is documented in an issue here https://github.com/golang/go/issues/20135

This code snippet below shows that the allocated memory is the same even when all the elements are deleted from the map:

```
package main

import (
        "fmt"
        "runtime"
)

func bToMb(b uint64) uint64 {
        return b / 1024 / 1024
}

func PrintMemUsage() {
        var m runtime.MemStats
        runtime.ReadMemStats(&m)
        // For info on each, see: https://golang.org/pkg/runtime/#MemStats
        fmt.Printf("Alloc = %5d MiB", bToMb(m.Alloc))
        fmt.Printf("\tTotalAlloc = %5d MiB", bToMb(m.TotalAlloc))
        fmt.Printf("\tSys = %5d MiB", bToMb(m.Sys))
        fmt.Printf("\tNumGC = %2d\n", m.NumGC)
}

func main() {
        myMap := make(map[int]uint64)

        fmt.Println("ADD")

        for i := 0; i < 100000000; i++ {
                myMap[i] = uint64(i + 1)
                if i%10000000 == 0 {
                        PrintMemUsage()
                }
        }
        PrintMemUsage()

        fmt.Println()
        fmt.Println("DELETE")

        for i := 0; i < 100000000; i++ {
                delete(myMap, i)
                if i%10000000 == 0 {
                        PrintMemUsage()
                }
        }

        fmt.Println("myMap len:", len(myMap))
        PrintMemUsage()

        myMap = nil
        fmt.Println()
        fmt.Println("myMap set to nil")

        fmt.Println("Manual gc trigger")
        runtime.GC()

        PrintMemUsage()
}
```

In our current code, there are 2 maps that are used extensively for the entire duration of the code. One is `positionMap` in forest and the other is `memttldb`.

For forest, a new const `maxMapCount` is declared along with a new forest member variable `mapFreeCount`. For each deletion from the `positionMap`, `mapFreeCount` is incremented once. In `forest.cleanup()`, if `mapFreeCount` is greater than `maxMapCount`, a new map is created, letting the old map be garbage collected.

For `memttldb`, a new map is created during every flush.

# Second change

For `utreexoserver`, the flag `-memttldb` needed to be given in order for `-allttldbinmem` to work. The new code now automatically turns on flag `-memttldb` when flag `-allttldbinmem` is given.

### DIFF
--- a/accumulator/forest.go
+++ b/accumulator/forest.go
@@ -11,6 +11,10 @@ import (
 const (
 	sibSwap       = false
 	bridgeVerbose = false
+
+	// maxMapCount denotes how many deletes from a map there can be
+	// before a new map is created.
+	maxMapCount = 4000000
 )
 
 // A FullForest is the entire accumulator of the UTXO set. This is
@@ -63,6 +67,9 @@ type Forest struct {
 
 	// map from hashes to positions.
 	positionMap map[MiniHash]uint64
+
+	// mapFreeCount keeps track of how many deletions happened from a map
+	mapFreeCount int
 
 	/*
 	 * below are just for testing / benchmarking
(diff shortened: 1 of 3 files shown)


kcalvinalvin:
I'll do some more extensive memory usage checks. I've only done a quick run on testnet so far and it does seem to limit memory usage since there are spikes in the utxo set here and there.

I'll convert to draft for now

kcalvinalvin:
closing since the UTXO set mostly grows over time. And there are other PRs that reduce/will reduce memory usage.

Issue #287, opened 2021-07-07
There's no way of telling why a polNode is supposed to be remembered


kcalvinalvin:
## The problem

To remember a polNode, we set the `niece[0]` of the polNode to itself. Telling whether a leaf is supposed to be remembered or not can be figured out by `polNode.niece[0] == polNode`.

There's two cases where a polNode needs to be remembered:

1: The leaf will be spent soon and is therefore cached.
2: The leaf is a sibling of the leaf to be deleted soon and is therefore cached. Important to note that the leaf by itself doesn't necessarily need to be cached as it'll stick around for a while.

These two cases need to be handled differently. Before decision tree shows how each case should be handled.
<img width="1216" alt="Screen Shot 2021-07-07 at 3 13 37 PM" src="https://user-images.githubusercontent.com/37185887/124708962-eac56d80-df35-11eb-8def-7fcbbe275786.png"> 

However, since there's only one way of remembering `polNode.niece[0] = polNode`, there's no way of doing this.

## Suggestion

Easiest way to solve this would be to denote case 2 as `polNode.niece[1] = polNode`. This might be a good stopgap until we have a swapless accumulator design working.

But in my opinion, I think maybe separating out the roots and leaves as different structs might work better. A `polNode` could be an interface and we can have `root`, `leaf` and `node` as this is how we're treating the polNodes anyways. This is a bigger change.

Issue #285, opened 2021-07-05
accumulator: PolNode remember doesn't work as intended


kcalvinalvin:
The way we currently have implemented the remember functionality is to have the `niece[0]` be itself. However, this property doesn't hold the way we currently have implemented `addOne()`.

https://github.com/mit-dci/utreexo/blob/3f55cb12f16dec7f9d8919b020a6f5d7bf636ced/accumulator/pollard.go#L151-L154

In this code snippet, we point the `niece[0]` to to itself to prevent the node from being pruned.

Example:

```
02
|---\
00  01
```

Here the leaf count is 2 and is `10` in binary. There is a 0 immediately so we can just simply append.

kcalvinalvin:
Mostly trying to get a `GetSize` function working for pollard for debugging so we can see how many bytes the pollard is taking up really.

```
func (n *polNode) GetSize(size int) int {
        // n.niece[0] is set to itself to be remembered in addOne()
        if n.niece[0] != nil {
                // hash and two pointers
                size += 32 + 16

                if n.niece[0] != n {
                        n.niece[0].GetSize(size)
                }
        }
        if n.niece[1] != nil {
                // hash and two pointers
                size += 32 + 16

                n.niece[1].GetSize(size)
        }

        return size
}
```
I think this function should work but since n.niece[0] isn't really n, it causes infinite loops for following `niece[0]` polNodes (since the leave will point back up).

dergoegge:
I actually think that the current approach works as intended but may lead to some extra memory usage is some cases.

> If `03` was to be remembered, `03.niece[0]` would be a pointer to `03`.

We want to remember the proof for a leaf but not necessarily the leaf itself. If `03` points to itself we would prune `02` which is part of the proof for `03`.

> `02.niece[0]` now points to `03`

In your example this is what we want, `02` pointing to `03` means that the proof for `03` will not be pruned.

Lets assume we have this
```
12
|-------\
08      09      10
|---\   |---\   |---\
00  01  02->03  04  05   06
```

**what happens if the proof for `03` (`02`, `08`) changes because `02` gets deleted?**

We swap `02` and `06` resulting in:
```
12
|-------\
08      09      10
|---\   |---\   |---\
00  01  06->03  04  05   02
```
now `02` is just dropped, resulting in the modified forest. Since now `06` still points to `03` we have what we want, the proof for `03` (`06`, `08`) is still cached.

**what happens if `03` itself moves?**
[...]

PR #288, opened 2021-07-07
Fix random polNodes sticking around

kcalvinalvin:
This PR is related with #287

EDIT:

Second commit adds pollard size to the stats, making it easier to keep track of the pollard size.

### DIFF
--- a/accumulator/pollard.go
+++ b/accumulator/pollard.go
@@ -148,13 +148,13 @@ func (p *Pollard) addOne(add Hash, remember bool) error {
 
 	n := new(polNode)
 	n.data = add
-	if remember || p.positionMap != nil {
-		// flag this leaf as memorable via it's left pointer
-		n.niece[0] = n // points to itself (mind blown)
-	}
+	n.remember = remember
 
 	if p.positionMap != nil {
 		p.positionMap[add.Mini()] = p.numLeaves
+
+		// Always remember everything for full pollard.
+		n.remember = true
 	}
 
 	// if add is forgetable, forget all the new nodes made
@@ -162,9 +162,8 @@ func (p *Pollard) addOne(add Hash, remember bool) error {
 	var h uint8
 	for ; (p.numLeaves>>h)&1 == 1; h++ {
 		// grab, pop, swap, hash, new
-		leftRoot := p.roots[len(p.roots)-1] // grab
-		p.roots = p.roots[:len(p.roots)-1]  // pop
-
+		leftRoot := p.roots[len(p.roots)-1]                        // grab
+		p.roots = p.roots[:len(p.roots)-1]                         // pop
 		leftRoot.niece, n.niece = n.niece, leftRoot.niece          // swap
 		nHash := parentHash(leftRoot.data, n.data)                 // hash
 		n = &polNode{data: nHash, niece: [2]*polNode{leftRoot, n}} // new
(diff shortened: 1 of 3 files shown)


adiabat:
try:

```
if h!=0 {
    leftRoot.niece, n.niece = n.niece, leftRoot.niece          // swap
}
```
and for ```n.prune()``` it can take an `h==0` bool as an argument, which if true gives this different functionality:
```
if n.niece[0].deadEnd() && n.niece[1].deadEnd() {
	n.niece[0], n.niece[1] = nil, nil
}
```

dergoegge:
The last part of my comment [here](https://github.com/mit-dci/utreexo/issues/285#issuecomment-891234843) might explain why there are random nodes sticking around.
I think the simplest solution to remembering proofs is to use a dedicated polnode that leaves point to if they are supposed to be remembered.

accumulator/pollardutil.go
```diff
@@ -71,6 +71,15 @@ func (n *polNode) prune() {
 	}
 }
 
+// getSize returns the size of the polNode and all the nieces below it in bytes.
+func getSize(n *polNode) int64 {
+	if n == nil {
+		return 0
+	}
+	// hash (32), two pointers (16), and a bool (1)
+	return (getSize(n.niece[0]) + 32 + 16 + 1 + getSize(n.niece[1]))
```
dergoegge, line 80: Using `unsafe.Sizeof` might be better here.
https://play.golang.org/p/fY3YmirNFSe
The Go compiler aligns the bool to 8 bytes.

accumulator/pollard.go
```diff
@@ -209,6 +208,23 @@ func (p *Pollard) rem2(dels []uint64) error {
 		}
 	}
 
+	// All the leaves to be deleted should be set to be not remembered.
+	// TODO there's probably a better way to do this than calling readPos
+	// a whole lot.
+	for _, del := range dels {
```
dergoegge, line 226: it would probably be better to set remember=false when swapping the leaves

accumulator/pollardutil.go
```diff
@@ -9,8 +9,9 @@ import (
 
 // PolNode is a node in the pollard forest
 type polNode struct {
-	data  Hash
-	niece [2]*polNode
+	data     Hash
+	niece    [2]*polNode
+	remember bool
```
dergoegge, line 14: maybe creating a `polLeaf` type would be good because the remember bool is stored on all nodes now but really only the leaves need it.
Both `polNode` and `polLeaf` would need to inherit from the same interface.

dergoegge (commented):
I feel like creating a special `rememberNode` would be the easier patch with less side effects.
`rememberNode` could either be a global or a field on `Pollard`.

accumulator/pollard.go
kcalvinalvin: I did try to get that to work but `TestCache()` kept failing and I bet there's a bunch of edge cases to handle. I'm leaning more towards just getting this in and then working on setting remember to false during swaps since that's an optimization.

accumulator/pollardutil.go
kcalvinalvin: Hmm yeah. I also hate the fact that `niece` isn't nieces for roots and leaves. Maybe we can solve all these problems at once.

accumulator/pollardutil.go
kcalvinalvin: Fixed.

adiabat:
OK - this fixes the bug but at the cost of increasing the memory usage around 17% (extra 8 bytes for the bool on all nodes).
We need to fix that space usage, but will merge this for now because it does fix the caching.
We might fix this by going to pointing to a (global) "memorable" dummy node to point to.  Or some other larger fix to pollard reprsentation.

Issue #295, opened 2021-07-19
Add tests for pollard serialize and deserialize


kcalvinalvin:
https://github.com/mit-dci/utreexo/blob/7d436ce11280c3f0e2fd7b4f56e43d012125b811/accumulator/pollardutil.go#L141-L188

We're currently lacking test for pollard serialize and deserialize. It'd be good to add tests for these so that we don't break things in the future.

We already have tests were we generate pollard to run other tests on (notably `TestPollardRand()` in pollard_test.go).

To test if serialize and deserialize indeed works how we expect them to, we can:

1. Serialize a pollard object
2. Save the bytes of this serialization
3. Deserialize into a pollard object
4. Re-serialize the pollard object into byte
5. Use `bytes.Compare()` to see if the byte serializations are equal

kcalvinalvin:
> Should the testing for the serialisation and de-serialisation go into the pollardRandomRemember function at the end (as a pollard is generated there) . If the bytes comparison at the step 5 as mentioned above is verified the return nil else return error since return type is error.

We can do that too. It'd be clearer if the serialization was separated out into its own test though. We also don't need other components of `pollardRandomRemember` so it'd be good to leave them out to keep the test speeds quick.

Generating a pollard isn't hard to do. We just have to add a slice of `Leaf` to a pollard object.

A new pollard object can be created in-memory just with a declaration: ```var pollard Pollard``` or ```pollard := new(Pollard)```.

Then we need a slice of `Leaf` (`[]Leaf`) to add to the pollard. https://github.com/mit-dci/utreexo/blob/7d436ce11280c3f0e2fd7b4f56e43d012125b811/accumulator/types.go#L51-L56

Leaf is just a Hash and a bool. Hash is just [32]byte. So we can create a leaf with something like
```
l := Leaf{
    // Generate some [32]byte slice. We can hash []byte{1} or anything else here.
    Hash: sha512.Sum512_256([]byte{1}),
    // Doesn't matter if it's really just true or false.
    Remember: true,
}
```

We generate a bunch of these `Leaf` and make a slice of them. 

Then we can just add a slice of `Leaf`s with `pollard.add(leafslice)`. Volia! We now have a pollard we can serialize and deserialize.

PR #296, opened 2021-07-21
Pollard test for serialize and deserailize 

DiptoChakrabarty:
This works on the issue of  #295 

function PollardSerializeDeserialize() generates a pollard and serializes it and then deserailizes it  keeping account of the bytes count of the previous serialization.

The desearialized pollard is serailized again and the bytes obtained is compared with the previous serialization bytes.
If equal the test passes .

function TestPollardSerializeDeserialize() performs the testing

### DIFF
--- a/accumulator/pollard_test.go
+++ b/accumulator/pollard_test.go
@@ -1,8 +1,6 @@
 package accumulator
 
 import (
-	"bytes"
-	"errors"
 	"fmt"
 	"math/rand"
 	"testing"
@@ -34,16 +32,6 @@ func TestPollardFixed(t *testing.T) {
 		t.Fatal(err)
 	}
 }
-
-func TestPollardSerializeDeserialize(t *testing.T) {
-	rand.Seed(2)
-
-	err := PollardSerializeDeserialize()
-	if err != nil {
-		t.Fatal(err)
-	}
-}
-
 func TestPollardSimpleIngest(t *testing.T) {
 	f := NewForest(nil, false, "", 0)
 	adds := make([]Leaf, 15)
@@ -221,52 +209,6 @@ func fixedPollard(leaves int32) error {
 
 	if !p.equalToForest(f) {
 		return fmt.Errorf("p != f (leaves)")
-	}
-
-	return nil
-}
-
-func PollardSerializeDeserialize() error {
-	var p Pollard
-
-	// generate slice of leaf
-	leaves := make([]Leaf, 10)
-
-	for i := 0; i < len(leaves); i++ {
-		leaves[i].Hash[0] = uint8(i + 1)
-	}
-
-	// add leaves to pollard
-	p.add(leaves)
-
-	// performing serialization
-
-	old_byte, err := p.Serialize()
-
-	if err != nil {
-		return err
-	}
-
-	// perform Deserialize
-
-	err = p.Deserialize(old_byte)
-	if err != nil {
-		return err
-	}
-
-	// Serialize again and compare bytes
-
-	new_byte, err := p.Serialize()
-
-	if err != nil {
-		return err
-	}
-
-	res := bytes.Compare(old_byte, new_byte)
-
-	// If comaprison unequal return error
-	if res != 0 {
-		return errors.New("Bytes Unequal")
 	}
 
 	return nil
--- a/accumulator/pollard_test.go
+++ b/accumulator/pollard_test.go
@@ -1,6 +1,8 @@
 package accumulator
 
 import (
+	"bytes"
+	"errors"
 	"fmt"
 	"math/rand"
 	"testing"
@@ -210,6 +212,51 @@ func fixedPollard(leaves int32) error {
 
 	if !p.equalToForest(f) {
 		return fmt.Errorf("p != f (leaves)")
+	}
+
+	return nil
+}
+
+func PollardSerializeDeserialize() error {
+	var p Pollard
+
+	// generate slice of leaf
+	leaves := make([]Leaf, 10)
+
+	for i := 0; i < len(leaves); i++ {
+		leaves[i].Hash[0] = uint8(i + 1)
+	}
+
+	// add leaves to pollard
+	p.add(leaves)
+
+	// performing serialization
+
+	old_byte, err := p.Serialize()
+
+	if err != nil {
+		return err
+	}
+
+	// perform Deserialize
+
+	err = p.Deserialize(old_byte)
+	if err != nil {
+		return err
+	}
+
+	// Serialize again and compare bytes
+
+	new_byte, err := p.Serialize()
+
+	if err != nil {
+		return err
+	}
+
+	res := bytes.Compare(old_byte, new_byte)
+
+	if res != 0 {
+		return errors.New("Bytes Unequal")
 	}
 
 	return nil
--- a/accumulator/pollard_test.go
+++ b/accumulator/pollard_test.go
@@ -35,6 +35,15 @@ func TestPollardFixed(t *testing.T) {
 	}
 }
 
+func TestPollardSerializeDeserialize(t *testing.T) {
+	rand.Seed(2)
+
+	err := PollardSerializeDeserialize()
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
 func TestPollardSimpleIngest(t *testing.T) {
 	f := NewForest(nil, false, "", 0)
 	adds := make([]Leaf, 15)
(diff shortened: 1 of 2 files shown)


accumulator/pollard_test.go
```diff
@@ -215,6 +226,52 @@ func fixedPollard(leaves int32) error {
...
+
+	for i := 0; i < len(leaves); i++ {
+		leaves[i].Hash[0] = uint8(i + 1)
+	}
+
+	// add leaves to pollard
+	p.add(leaves)
+
+	// performing serialization
+
```
kcalvinalvin, line 243: Don't need the empty line here

accumulator/pollard_test.go
```diff
@@ -215,6 +226,52 @@ func fixedPollard(leaves int32) error {
...
+	old_byte, err := p.Serialize()
+
```
kcalvinalvin, line 245: Don't need the empty line here

accumulator/pollard_test.go
```diff
@@ -215,6 +226,52 @@ func fixedPollard(leaves int32) error {
...
+	if err != nil {
+		return err
+	}
+
+	// perform Deserialize
+
```
kcalvinalvin, line 251: Don't need the empty line here

accumulator/pollard_test.go
```diff
@@ -215,6 +226,52 @@ func fixedPollard(leaves int32) error {
...
+	err = p.Deserialize(old_byte)
+	if err != nil {
+		return err
+	}
+
+	// Serialize again and compare bytes
+
```
kcalvinalvin, line 258: Don't need the empty line here

accumulator/pollard_test.go
```diff
@@ -215,6 +226,52 @@ func fixedPollard(leaves int32) error {
...
+	new_byte, err := p.Serialize()
+
```
kcalvinalvin, line 260: Don't need the empty line here

accumulator/pollard_test.go
```diff
@@ -215,6 +226,52 @@ func fixedPollard(leaves int32) error {
...
+	if err != nil {
+		return err
+	}
+
+	res := bytes.Compare(old_byte, new_byte)
```
kcalvinalvin, line 265: There's a `bytes.Equal()` that outputs a bool (true if equal, false if not equal). Would be better to use that instead of `bytes.Compare()`.

accumulator/pollard_test.go
```diff
@@ -215,6 +226,52 @@ func fixedPollard(leaves int32) error {
...
+
+	// add leaves to pollard
+	p.add(leaves)
+
+	// performing serialization
+
+	old_byte, err := p.Serialize()
+
+	if err != nil {
+		return err
```
kcalvinalvin, line 247: After changing the function to `TestPollardSerializeDeserialize(t *testing.T)`, this should be changed to `t.Fatal(err)` since Golang testing function signature must be `func TestXXX(t *testing.T) {}`

`t.Error()` (or `t.Errorf()` to have formatting) for errors that aren't fatal and when the test can still go on to give more useful information for debugging. `t.Fatal()` for errors that force the test to be stopped.

kcalvinalvin (changes requested):

kcalvinalvin:
Also, the test should be moved to a separate file called `pollardutil_test.go` since the `Pollard.Serialize()` and `Pollard.Deserialize()` functions are in the file `pollardutil.go`

accumulator/pollard_test.go
```diff
@@ -215,6 +226,52 @@ func fixedPollard(leaves int32) error {
...
+	}
+
+	// perform Deserialize
+
+	err = p.Deserialize(old_byte)
```
kcalvinalvin, line 252: Let's make a new pollard object here and deserialize into that.

accumulator/pollard_test.go
```diff
@@ -33,6 +35,15 @@ func TestPollardFixed(t *testing.T) {
 	}
 }
 
+func TestPollardSerializeDeserialize(t *testing.T) {
```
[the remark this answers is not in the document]
DiptoChakrabarty: okay just to confirm everything of PollardSerializeDeserialize can be directly written to the TestPollardSerializeDeserialize(t *testing.T)  in a new pollardutil_test  file

accumulator/pollard_test.go
DiptoChakrabarty: okay working on it

accumulator/pollard_test.go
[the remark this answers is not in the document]
kcalvinalvin: >  everything of PollardSerializeDeserialize can be directly written to the TestPollardSerializeDeserialize(t *testing.T) 
Yes.

>in a new pollardutil_test file
Yes. The exact name would be `pollardutil_test.go`

accumulator/pollardutil_test.go
```diff
@@ -0,0 +1,45 @@
...
+	"bytes"
+	"testing"
+)
+
+func TestPollardSerializeDeserialize(t *testing.T) {
+	var p, q Pollard
+	// generate slice of leaf
+	leaves := make([]Leaf, 10)
+	for i := 0; i < len(leaves); i++ {
+		leaves[i].Hash[0] = uint8(i + 1)
```
adiabat, line 13: One thing to watch out for: there are parts of the code that treat an all 0 hash as invalid / uninitialized / empty.  To avoid hitting those checks, either start the loop at 1, or add something like `leaves[i].Hash[1] = 0xff`

accumulator/pollardutil_test.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	// add leaves to pollard
+	err := p.add(leaves)
+	if err != nil {
+		t.Fatal(err)
+	}
+	// new pollard
+	err = q.add(leaves)
```
adiabat, line 21: Why add leaves to q, as you then deserialize into q?
Actually maybe this should lead to an error if you try to deserialize into an already initialized / filled pollard.

adiabat (commented):
2 things to look at

accumulator/pollardutil_test.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	// performing serialization
+	old_byte, err := p.Serialize()
+	if err != nil {
+		t.Fatal(err)
+	}
+	// perform Deserialize
+	err = q.Deserialize(old_byte)
```
adiabat, line 31: Actually, the more I think about it, this call to `Deserialize` should throw an error.  The fact that it doesn't right now seems like a bug.

The reason is: this test is supposed to test serialize / deserialize.  But if `Deserialize` did absolutely nothing, this test would still pass.  Because p and q already have the same hashes in them.

I think a check like 
```if numleaves != 0 {
return fmt.Errorf("can't deserialize into populated pollard")
}
```
would fix this and avoid the problem.

accumulator/pollardutil_test.go
DiptoChakrabarty: I am using i+1 for each so it starts from 1  , so there should be no 0 hash right. I can  still go ahead with starting with 1.

accumulator/pollardutil_test.go
DiptoChakrabarty: So  should I first serialize q then perform deserailization and then compare .

accumulator/pollard_test.go
```diff
@@ -32,7 +32,6 @@ func TestPollardFixed(t *testing.T) {
 		t.Fatal(err)
 	}
 }
-
```
kcalvinalvin, line 35: Unnecessary line deletion here.

accumulator/pollardutil_test.go
kcalvinalvin: Yup I don't see the all 0 issue here

accumulator/pollardutil_test.go
kcalvinalvin: The test should check if the serialized bytes indeed deserialize correctly from the bytes. So there should be no `q.add()` since this will negate the deserialization test.

If we do the `q.add()`, the test will pass even if `err = q.Deserialize(old_byte)` fails.

adiabat:
it's tricky - my default assumption is that if you have a struct with everything already filled in, and call `.Deserialize()` on it, there should be an error.  But it looks like everywhere else in btcd if you do something like this, it just overwrites everything in the struct from the deserialized data.

If we stick with the btcd way of doing things, this test should first *not* insert anything into `q` and make sure it's different from `p`, then `q.Deserialize()`, then check that they're the same.  This would mean we don't have to change anything about `Deserialize()` itself.

Issue #294, opened 2021-07-18
Use Enums for forest types


kcalvinalvin:
We have 4 different types of forest and the way to specify each one of them is with an argument. It'd be a lot cleaner to implement these as golang enums (example - yourbasic.org/golang/iota/) instead of having many function arguments.

https://github.com/mit-dci/utreexo/blob/7d436ce11280c3f0e2fd7b4f56e43d012125b811/accumulator/forest.go#L96-L117

We'd have to change this function and then make changes to all the functions that call `NewForest()`.

rohunmsharman:
Is this what you meant? I am a little unsure as this only removes one function argument. 
<img width="845" alt="enumForestType" src="https://user-images.githubusercontent.com/66086705/126711049-abece061-4e34-4ce1-ae21-c208ed92cdcf.png">

r-ush:
@kcalvinalvin I have sent a PR that solves this issue. 😃

kcalvinalvin:
> Is this what you meant? I am a little unsure as this only removes one function argument.

Yes that is what I meant. Though the `NewForest()` function should not be a method on the type.

PR #297, opened 2021-07-23
ref: NewForest function to use enums for forest types

r-ush:
Solves issue #294
Uses an enum as a receiver to the function `NewForest`

```go
const (
	DiskForest ForestName = iota
	RamForest
	CacheForest
	CowForest
)
```
Earlier the new forest function used function arguments to infer which type of forest is to be initialized
```go
if forestFile == nil {
		if cowPath == "" {
			// for in-ram
			f.data = new(ramForestData)
		} else {
			// Init cowForest
			d, err := initialize(cowPath, cowMaxCache)
			if err != nil {
				panic(err)
			}
			f.data = d
		}

	} else {
		// forest on disk or cached
		if cached {
			d := new(cacheForestData)
			d.file = forestFile
			d.cache = newDiskForestCache(20)
			f.data = d
		} else {
			// for on-disk
			d := new(diskForestData)
			d.file = forestFile
			f.data = d
		}
	}
```

Now the implementation has been made like this
```go
	switch fn {
	case DiskForest:
		d := new(diskForestData)
		d.file = forestFile
		f.data = d
	case RamForest:
		f.data = new(ramForestData)
	case CacheForest:
		d := new(cacheForestData)
		d.file = forestFile
		d.cache = newDiskForestCache(20)
		f.data = d
	case CowForest:
		d, err := initialize(cowPath, cowMaxCache)
		if err != nil {
			panic(err)
		}
		f.data = d
	}
```

### DIFF
--- a/accumulator/forest.go
+++ b/accumulator/forest.go
@@ -93,59 +93,57 @@ type Forest struct {
 	timeInVerify time.Duration
 }
 
+// ForestType defines the 4 type of forests:
+// DiskForest, RamForest, CacheForest, CowForest
+type ForestType int
+
+const (
+	// DiskForest  - keeps the entire forest on disk as a flat file. Is the slowest
+	//               of them all. Pass an os.File as forestFile to create a DiskForest.
+	DiskForest ForestType = iota
+	// RamForest   - keeps the entire forest on ram as a slice. Is the fastest but
+	//               takes up a lot of ram. Is compatible with DiskForest (as in you
+	//               can restart as RamForest even if you created a DiskForest. Pass
+	//               nil, as the forestFile to create a RamForest.
+	RamForest
+	// CacheForest - keeps the entire forest on disk but caches recent nodes. It's
+	//               faster than disk. Is compatible with the above two forest types.
+	//               Pass cached = true to create a cacheForest.
+	CacheForest
+	// CowForest   - A copy-on-write (really a redirect on write) forest. It strikes
+	//               a balance between ram usage and speed. Not compatible with other
+	//               forest types though (meaning there isn't functionality implemented
+	//               to convert a CowForest to DiskForest and vise-versa). Pass a filepath
+	//               and cowMaxCache(how much MB to use in ram) to create a CowForest.
+	CowForest
+)
+
 // NewForest initializes a Forest and returns it. The given arguments determine
-// what type of forest it will be. There are currently four types:
-//
-// DiskForest  - keeps the entire forest on disk as a flat file. Is the slowest
-//               of them all. Pass an os.File as forestFile to create a DiskForest.
-//
-// RamForest   - keeps the entire forest on ram as a slice. Is the fastest but
-//               takes up a lot of ram. Is compatible with DiskForest (as in you
-//               can restart as RamForest even if you created a DiskForest. Pass
-//               nil, as the forestFile to create a RamForest.
-//
-// CacheForest - keeps the entire forest on disk but caches recent nodes. It's
-//               faster than disk. Is compatible with the above two forest types.
-//               Pass cached = true to create a cacheForest.
-//
-// CowForest   - A copy-on-write (really a redirect on write) forest. It strikes
-//               a balance between ram usage and speed. Not compatible with other
-//               forest types though (meaning there isn't functionality implemented
-//               to convert a CowForest to DiskForest and vise-versa). Pass a filepath
-//               and cowMaxCache(how much MB to use in ram) to create a CowForest.
-func NewForest(forestFile *os.File, cached bool,
-	cowPath string, cowMaxCache int) *Forest {
+// what type of forest it will be.
+func NewForest(forestType ForestType, forestFile *os.File, cowPath string, cowMaxCache int) *Forest {
 
 	f := new(Forest)
 	f.numLeaves = 0
 	f.rows = 0
 
-	if forestFile == nil {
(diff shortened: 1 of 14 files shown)


accumulator/forest.go
```diff
@@ -113,39 +113,40 @@ type Forest struct {
 //               forest types though (meaning there isn't functionality implemented
 //               to convert a CowForest to DiskForest and vise-versa). Pass a filepath
 //               and cowMaxCache(how much MB to use in ram) to create a CowForest.
-func NewForest(forestFile *os.File, cached bool,
-	cowPath string, cowMaxCache int) *Forest {
+
+type ForestName int
```
kcalvinalvin, line 117: ForestName isn't quite right. There are different types of forests so this should be forestType

accumulator/forest.go
```diff
@@ -113,39 +113,40 @@ type Forest struct {
...
+
+const (
```
kcalvinalvin, line 119: For all the forest types here, we should copy over the comment from the `NewForest()` function and place it here.

```
  // DiskForest  - keeps the entire forest on disk as a flat file. Is the slowest
  //               of them all. Pass an os.File as forestFile to create a DiskForest.
  //
  // RamForest   - keeps the entire forest on ram as a slice. Is the fastest but
  //               takes up a lot of ram. Is compatible with DiskForest (as in you
  //               can restart as RamForest even if you created a DiskForest. Pass
  //               nil, as the forestFile to create a RamForest.
  //
  // CacheForest - keeps the entire forest on disk but caches recent nodes. It's
  //               faster than disk. Is compatible with the above two forest types.
  //               Pass cached = true to create a cacheForest.
  //
  // CowForest   - A copy-on-write (really a redirect on write) forest. It strikes
  //               a balance between ram usage and speed. Not compatible with other
  //               forest types though (meaning there isn't functionality implemented
  //               to convert a CowForest to DiskForest and vise-versa). Pass a filepath
  //               and cowMaxCache(how much MB to use in ram) to create a CowForest.
```

accumulator/forest.go
```diff
@@ -113,39 +113,40 @@ type Forest struct {
...
+	DiskForest ForestName = iota
+	RamForest
+	CacheForest
+	CowForest
+)
+
+func (fn ForestName) NewForest(forestFile *os.File, cowPath string, cowMaxCache int) *Forest {
```
kcalvinalvin, line 126: All exported functions in Go should have a comment with the format:

```
// FunctionName is ...
```

This is how go documentation is done. The documentation is automatically parsed and for package `accumulator`, it is at https://pkg.go.dev/github.com/mit-dci/utreexo/accumulator

kcalvinalvin: Also, the function `NewForest()` should not be a method on `ForestName`. It should accept a `ForestName` (`ForestName` should be renamed to `ForestType`) value as a function argument.

accumulator/readme.md
```diff
@@ -1,28 +1,25 @@
 A dynamic hash-based accumulator.
 
-Overview
---------
+## Overview
```
kcalvinalvin, line 3: There's no need for these changes since github renders the existing ones just fine right? (or are there IDEs that don't render these right?)

kcalvinalvin (changes requested):

accumulator/readme.md
r-ush: My code editor autoformatted it to a `##`, will revert it back in my next commit.

accumulator/forest.go
```diff
@@ -93,59 +93,61 @@ type Forest struct {
...
+	RamForest
+	// CacheForest - keeps the entire forest on disk but caches recent nodes. It's
+	//               faster than disk. Is compatible with the above two forest types.
+	//               Pass cached = true to create a cacheForest.
+	CacheForest
+	// CowForest   - A copy-on-write (really a redirect on write) forest. It strikes
+	//               a balance between ram usage and speed. Not compatible with other
+	//               forest types though (meaning there isn't functionality implemented
+	//               to convert a CowForest to DiskForest and vise-versa). Pass a filepath
+	//               and cowMaxCache(how much MB to use in ram) to create a CowForest.CowForest   - A copy-on-write (really a redirect on write) forest. It strikes
```
kcalvinalvin, line 117: I think there was a copy & paste error here

kcalvinalvin:
Besides the copy&paste error, I think it's good. I'm actually now thinking maybe the `NewForest()` function could just create os.File itself and not require a file as an argument... (but this would be another PR)

r-ush:
Fixed and squashed @kcalvinalvin 🚀

accumulator/forest.go
```diff
@@ -93,59 +93,57 @@ type Forest struct {
 	timeInVerify time.Duration
 }
 
+// forestType defines the 4 type of forests:
+// DiskForest, RamForest, CacheForest, CowForest
+type forestType int
```
kcalvinalvin, line 98: forestType should be capitalized since it needs to be accessible by callers that are importing package `accumulator`. So at the moment I don't think `cmd/utreexoserver` and `cmd/utreexoclient` builds (I've found other build errors but that's because of some dependency thing we have).

r-ush: For `cmd/utreexoserver` to build we will have to change the occurrences of `NewForest` function in utcd to accept the new arguments.

For now should I just change the `type forestType int` to `type ForestType int`?

PR #298, opened 2021-07-26
util/util, csn, bridgenode, wire: Use local DedupeBlock()

kcalvinalvin:
Use local DedupeBlock() to remove utcutil and utcd dependency. The
utreexo repository is now only dependent on the default btcd repository.

DedupeBlock() now also returns the count of all the inputs and outputs.
This let's callers pre-allocate slices to the required length improving
memory allcation efficiency.

## Note

This will make the `utreexoclient` and `utreexoserver` binaries a bit slower since they are now linked to the btcd repo without all the performance improvements I have added in. The eventual plan would be to link it to utcd, when things are a bit more stable.

### DIFF
--- a/bridgenode/processblock.go
+++ b/bridgenode/processblock.go
@@ -108,23 +108,24 @@ type txoStart struct {
 func blockToAddDel(bnr BlockAndRev) (
 	blockAdds []accumulator.Leaf, delLeaves []btcacc.LeafData, err error) {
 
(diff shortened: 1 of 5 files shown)


adiabat:
OK looks good.  Also hopefully `utreexoclient` and `utreexoserver` won't have to use too much from btcd anyway :)

PR #305, opened 2021-08-05
feature for saving undoBlocks (IN Progress)

DiptoChakrabarty:
This PR is for saving the undoblocks data returned from forest.Modify  from the BuildProofs function in the genproofs.go file.

A new channel of undo unBlock type has been created where the undoblocks are passed which is then sent to flatfileworker .

Undoblocks a method is added to serialize them into bytes.

ToDo -> hanlde case to save the undoblocks within the flatfileworker function


DiptoChakrabarty:
done by #306  , hence closing

PR #306, opened 2021-08-05
accumulator/undo: Add serialization for undoblocks

kcalvinalvin:
UndoBlocks need to be saved and re-accessed whenever there is a reorg
with the Bitcoin blockchain. Serialization and Deserialization is
added so that these blocks can be saved and fetched from disk.

UndoBlocks are also now exported so that callers will be able to access
the new Serialize() and Deserialize() methods.


accumulator/undo.go
```diff
@@ -34,8 +36,105 @@ func (u *undoBlock) ToString() string {
...
+		return err
+	}
+
+	for _, pos := range u.positions {
+		err := binary.Write(w, binary.BigEndian, pos)
+		if err != nil {
+			return err
+		}
+
+	}
```
dergoegge, line 68: It should be possible to do this:
```suggestion
	err = binary.Write(w, binary.BigEndian, u.positions)
	if err != nil {
		return err
	}
```

https://github.com/golang/go/blob/6e738868a7a943d7d4fd6bb1963e7f6d78111726/src/encoding/binary/binary.go#L354

accumulator/undo.go
```diff
@@ -34,8 +36,105 @@ func (u *undoBlock) ToString() string {
...
+
+	var posCount uint64
+	err = binary.Read(r, binary.BigEndian, &posCount)
+	if err != nil {
+		return err
+	}
+	u.positions = make([]uint64, posCount)
+
+	var pos uint64
+	for i := uint64(0); i < posCount; i++ {
```
dergoegge, line 105: Same as above just for `binary.Read`

accumulator/undo.go
kcalvinalvin: Change applied

accumulator/undo.go
kcalvinalvin: Change applied

PR #309, opened 2021-08-15
csn/ibd: Don't set UTXOs to be remembered

kcalvinalvin:
UTXOs are currently marked with ttl value of 0. Since 0 < 1000 is true,
the pollard was remembering all the UTXOs ever, leading to a ton of
memory usage. With this change, CSNs see a lot less memory usage.

This PR still sees the remembered leaves growing over time
so there might still be a problem yet to be fixed.

### DIFF
--- a/csn/ibd.go
+++ b/csn/ibd.go
@@ -192,8 +192,12 @@ func (c *Csn) putBlockInPollard(
 
 	remember := make([]bool, len(ub.UtreexoData.TxoTTLs))
 	for i, ttl := range ub.UtreexoData.TxoTTLs {
-		// ttl-ub.Height is the number of blocks until the block is spend.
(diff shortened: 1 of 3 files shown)


dergoegge:
Does the size of the pollard grow even if ttls are not used at all?

kcalvinalvin:
> Does the size of the pollard grow even if ttls are not used at all?

We still remember them so it does take up space. You can test it out on the master branch by commenting out this code
https://github.com/mit-dci/utreexo/blob/0c940de7b6d9ff864e8a1d1282298e7cbc8f7202/accumulator/pollard.go#L151-L154

With it commented out, the CSN barely uses any memory. With it, CSNs go up to 7GB around block 1,300,000 on testnet3

PR #315, opened 2021-08-21
Saving TTL data seperately

DiptoChakrabarty:
In this PR a new config for ttl directory has been added and ttl data are being stored in a new flatfilestate in the flatfileworker.go

### DIFF
--- a/bridgenode/flatfileworker.go
+++ b/bridgenode/flatfileworker.go
@@ -151,14 +151,6 @@ func flatFileWorker(
 				panic(err)
 			}
 		case ttlRes := <-ttlResultChan:
-			for ttlRes.Height > ff.currentHeight {
-				ud := <-proofChan
-				err = ff.writeProofBlock(ud)
-				if err != nil {
-					panic(err)
-				}
-			}
-
 			err = tf.writeTTLs(ttlRes)
 			if err != nil {
 				panic(err)
--- a/bridgenode/flatfileworker.go
+++ b/bridgenode/flatfileworker.go
@@ -58,11 +58,9 @@ type flatFileState struct {
 // pFileWorker takes in blockproof and height information from the channel
 // and writes to disk. MUST NOT have more than one worker as the proofs need to be
 // in order
-func flatFileWorker(
+
+func flatFileWorkerProofBlocks(
 	proofChan chan btcacc.UData,
-	ttlResultChan chan ttlResultBlock,
-	undoChan chan accumulator.UndoBlock,
-	leafblockChan chan int,
 	utreeDir utreeDir,
 	fileWait *sync.WaitGroup) {
 
@@ -88,8 +86,21 @@ func flatFileWorker(
 		panic(err)
 	}
 
-	// for the undofiles
+	ud := <-proofChan
+	err = ff.writeProofBlock(ud)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func flatFileWorkerUndoBlocks(
+	undoChan chan accumulator.UndoBlock,
+	utreeDir utreeDir,
+	fileWait *sync.WaitGroup) {
+
 	var uf flatFileState
+	var err error
+
 	uf.offsetFile, err = os.OpenFile(
 		utreeDir.UndoDir.offsetFile, os.O_CREATE|os.O_RDWR, 0600)
 	if err != nil {
@@ -108,8 +119,22 @@ func flatFileWorker(
 	if err != nil {
 		panic(err)
 	}
-	// ttl data saving initialization
+
+	undo := <-undoChan
+	err = uf.writeUndoBlock(undo)
+	if err != nil {
+		panic(err)
+	}
+}
+
+func flatFileWorkerTTlBlocks(
+	ttlResultChan chan ttlResultBlock,
+	leafblockChan chan int,
+	utreeDir utreeDir,
+	fileWait *sync.WaitGroup) {
+
 	var tf flatFileState
+	var err error
 
 	tf.offsetFile, err = os.OpenFile(
 		utreeDir.TtlDir.OffsetFile, os.O_CREATE|os.O_RDWR, 0600)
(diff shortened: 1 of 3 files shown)


kcalvinalvin:
> I am not getting any error like the index one , what I am getting is this
> Selection_127

Ah this is a separate error I think. If you delete `pollardFile` in your data directory, I bet it won't have that error.

> the server running is like this , this does not produce any error
> Selection_128

I think it's because it's already synced up it's not producing any errors. If you delete the directories with the data and restart, I think you'll come across the same error.

bridgenode/flatfileworker.go
```diff
@@ -312,23 +325,45 @@ func (ff *flatFileState) writeProofBlock(ud btcacc.UData) error {
...
+		if err != nil {
+			return err
+		}
+
+		// increment value of offset after before write operation by 4
+		ff.currentOffset = ff.currentOffset + 4
 
 		// write it's lifespan as a 4 byte int32 (bit of a waste as
 		// 2 or 3 bytes would work)
 		// add 16: 4 for magic, 4 for size, 4 for height, 4 numTTL, then ttls start
```
kcalvinalvin, line 357: Stale comment here. The 16 is gone now

bridgenode/flatfileworker.go
```diff
@@ -312,23 +325,45 @@ func (ff *flatFileState) writeProofBlock(ud btcacc.UData) error {
...
+		return err
+	}
+
 	// for all the TTLs, seek and overwrite the empty values there
 	for _, c := range ttlRes.Created {
 		// seek to the location of that txo's ttl value in the proof file
 
 		binary.BigEndian.PutUint32(
 			ttlArr[:], uint32(ttlRes.Height-c.createHeight))
+		// for each iteration of  append 0x00
```
kcalvinalvin, line 346: This code doesn't work since there should be space allocated *for every leaf*.

Try inserting this snippet at line 354 (also change line 120 to  `utreeDir.TtlDir.ttlsetFile, os.O_CREATE|os.O_RDWR, 0600)`

```
                 shouldBeZeros := make([]byte, 4)                                                                                                                                                                                              
                 _, err = ff.proofFile.ReadAt(shouldBeZeros, ff.offsets[c.createHeight]+int64(c.indexWithinBlock*4))                                                                                                                           
                 if err != nil {                                                                                                                                                                                                               
                         return err                                                                                                                                                                                                            
                 }                                                                                                                                                                                                                             
                 if !bytes.Equal(shouldBeZeros, []byte{0x00, 0x00, 0x00, 0x00}) {                                                                                                                                                              
                         err := fmt.Errorf("read %x, should have read all zeros", shouldBeZeros)                                                                                                                                               
                         panic(err)                                                                                                                                                                                                            
                 }
```

This will fail and say that you read some none 0s.

Per every leaf created, we need to allocate 0s on disk. This means that we need to allocate when we generate the leaf to add. (found here:)

https://github.com/mit-dci/utreexo/blob/5f26803e7056a7a8c1bae3414f1711f37518c3ba/bridgenode/genproofs.go#L145

So for each len(blockAdds), we need to

```
bytesToWrite := make([]byte, len(blockAdds))
_, err := ff.proofFile.WriteAt(bytesToWrite, ff.currentOffset)
if err != nil {
    return err
}
```

I think the best way to give this information to flatFileWorker may just be to have another channel for just sending over the blockAdds and having another function that handles just this allocation on disk.

bridgenode/flatfileworker.go
```diff
@@ -312,23 +325,45 @@ func (ff *flatFileState) writeProofBlock(ud btcacc.UData) error {
...
+		_, err := ff.proofFile.WriteAt([]byte{0x00, 0x00, 0x00, 0x00}, ff.currentOffset)
+		if err != nil {
+			return err
+		}
+
+		// increment value of offset after before write operation by 4
+		ff.currentOffset = ff.currentOffset + 4
```
kcalvinalvin, line 353: There's no reason to do this every loop. We can do

```
ff.currentOffset += len(ttlRes.Created) * 4
```

at the end of the loop.

kcalvinalvin (changes requested):

bridgenode/flatfileworker.go
DiptoChakrabarty: Okay Ill change it

bridgenode/flatfileworker.go
DiptoChakrabarty: Okay removing it

bridgenode/flatfileworker.go
```diff
@@ -147,6 +161,12 @@ func flatFileWorker(
 			if err != nil {
 				panic(err)
 			}
+		case size := <-leafblockChan:
+			bytesTtlWrite := make([]byte, size)
```
kcalvinalvin, line 165: Should be a filewait.Done() here

bridgenode/flatfileworker.go
```diff
@@ -107,47 +119,54 @@ func flatFileWorker(
...
+	}
+	tf.fileWait = fileWait
+
+	err = tf.ffInit()
+	if err != nil {
+		panic(err)
+	}
+
+	size := <-leafblockChan
+	bytesTtlWrite := make([]byte, size)
```
kcalvinalvin, line 158: `size` should be `size * 4`. This is because `size` indicates the count of additions and we write 4 bytes per addition.

Maybe might be better to do:

```
addCount := <-leafblockChan
size := addCount * 4
```
so that it's clearer what variable does whatl.

bridgenode/flatfileworker.go
```diff
@@ -107,47 +119,54 @@ func flatFileWorker(
...
+	_, err = tf.proofFile.WriteAt(bytesTtlWrite, tf.currentOffset)
+	tf.fileWait.Done()
+	if err != nil {
+		panic(err)
+	}
+	tf.currentOffset = tf.currentOffset + int64(size)
```
kcalvinalvin, line 164: Two things to change here:

1: Again, size should be `size * 4`. If we use:

```
addCount := <-leafblockChan
size := addCount * 4
```

then we can keep this line the same.

2: In `writeTTLs()` we have this line which indicates the start of the offset for the current block.
```
ff.offsets = append(ff.offsets, ff.currentOffset)
```

If we increment before doing that append, we're effectively pointing at the next block. This entire line should happen after the append. So we need to move this line after `writeTTLs()`, (which would be at line 170).

bridgenode/flatfileworker.go
```diff
@@ -107,47 +119,54 @@ func flatFileWorker(
...
+	if err != nil {
+		panic(err)
+	}
+	tf.fileWait = fileWait
+
+	err = tf.ffInit()
+	if err != nil {
+		panic(err)
+	}
+
```
kcalvinalvin, line 156: We need to add a `for{}` loop for this segment here. Otherwise, the program will do this function exactly once and exit (which is not what we want).

```
for {          
                size := <-leafblockChan
                bytesTtlWrite := make([]byte, size*4)
                _, err = tf.proofFile.WriteAt(bytesTtlWrite, tf.currentOffset)                                          
                tf.fileWait.Done()                                                                                      
                if err != nil {                                                                                         
                        panic(err)
                }         
                                                            
                ttlRes := <-ttlResultChan                   
                err = tf.writeTTLs(ttlRes)               
                if err != nil {
                        panic(err)                       
                }                                                                                                       
                                                            
                tf.currentOffset = tf.currentOffset + int64(size)*4
        }
```

We need something like the code above.

bridgenode/flatfileworker.go
```diff
@@ -87,8 +86,21 @@ func flatFileWorker(
 		panic(err)
 	}
 
-	// for the undofiles
+	ud := <-proofChan
```
kcalvinalvin, line 89: We need a `for{}` loop here as the program will execute this function exactly once and then exit.

```
for {                                               
                ud := <-proofChan                           
                err = ff.writeProofBlock(ud)             
                if err != nil {
                        panic(err)                       
                }                                                                                                       
        }
```

Something like the code above.

bridgenode/flatfileworker.go
```diff
@@ -107,47 +119,54 @@ func flatFileWorker(
...
-			if err != nil {
-				panic(err)
-			}
-		case undo := <-undoChan:
-			err = uf.writeUndoBlock(undo)
-			if err != nil {
-				panic(err)
-			}
-		}
+	undo := <-undoChan
```
kcalvinalvin, line 123: We also need a `for{}` loop here as the program will execute this function exactly once and then exit (which is not what we want).

```
for {                                                                                                           
                undo := <-undoChan                                                                                      
                err = uf.writeUndoBlock(undo)
                if err != nil {                                                                                         
                        panic(err)        
                }                                                                                                       
        }
```

We need something like the code above.

kcalvinalvin (changes requested):

bridgenode/flatfileworker.go
```diff
@@ -107,48 +121,62 @@ func flatFileWorker(
...
+		tf.fileWait.Done()
+		if err != nil {
+			panic(err)
 		}
+		ttlRes := <-ttlResultChan
+		err = tf.writeTTLs(ttlRes)
+		if err != nil {
+			panic(err)
+		}
+		tf.currentOffset = tf.currentOffset + int64(size*4)
```
kcalvinalvin, line 175: We're still incrementing before the append. We should increment after the append in line 177. 

Refer to this comment as to why - https://github.com/mit-dci/utreexo/pull/315#discussion_r702634052

PR #318, opened 2021-09-03
accumulator/pollard: Add a VerifyBatchProof method

kcalvinalvin:
Pollard now has a method just for verifying a given proof. This new
method does not alter the Pollard in any way, making it useful during
verifying mempool Bitcoin transactions.

Am looking for inputs in this because I'm thinking maybe we should take out the verification part from `IngestBatchProof()` and just have the caller call `Pollard.VerifyBatchProof()` for verifying, and call something like

```
// This function would force us to export MiniTree and Node, so might not be worth it.
func (p *Pollard) IngestBatchProof(trees []MiniTree, rootCandidates []Node) error {}
```

### DIFF
--- a/accumulator/pollardproof.go
+++ b/accumulator/pollardproof.go
@@ -4,8 +4,39 @@ import (
 	"fmt"
 )
 
+// VerifyBatchProof verifies the hash and the proof passed in. It does not
+// make any modifications to the pollard.
+//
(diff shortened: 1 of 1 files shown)


adiabat:
Works for now; we need to reword the ingestion / verification process

PR #321, opened 2021-09-09
github workflows: add coveralls for test coverage reports

### LINKED ISSUE #243: Test coverage
It would be nice to see which parts of the code are covered by the unit tests. 
Go has the option to generate coverage reports: 
`go test -coverprofile=cover.out`
Which can be viewed with:
`go tool cover -html=cover.out`

r-ush:
closes #243 

## what it does
Add [goveralls](https://github.com/mattn/goveralls) that gives reports on test coverage of the project.

## how it does
Currently covering package `accumulator`.

It gives reports like-->
image
image

I tried adding it to the `go.yml` workflow, but the Job fails with: Error: The operation was cancelled; this is probably because of the machine running out of memory when the tests are run with
```shell
go test -v ./accumulator -race -covermode atomic -coverprofile=covprofile
```
instead of originally
```shell
go test -v ./accumulator
```

## Need suggestion on
Is there a better way to structure the workflows? Because right now, it runs the unit tests twice, which is not ideal.


r-ush:
@kcalvinalvin @adiabat Please take a look at it 😄

.github/workflows/go.yml
```diff
@@ -4,119 +4,113 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
-    timeout-minutes: 5
```
kcalvinalvin, line 7: Is there a reason for this timeout increase?

kcalvinalvin:
Most of the code is the same right? It's a bit hard to review since there's removed lines that turn this into a decently sized change. Could you add those back in to minimize the lines of change?

.github/workflows/go.yml
r-ush: It is taking around 7 minutes to run the unit tests with the added flags, so it was getting timed out at 5 minutes. That is why it is increased

r-ush:
> Most of the code is the same right? It's a bit hard to review since there's removed lines that turn this into a decently sized change. Could you add those back in to minimize the lines of change?

Done!

.github/workflows/go.yml
```diff
@@ -74,6 +80,7 @@ jobs:
     needs: Build
     timeout-minutes: 5
     steps:
+
```
kcalvinalvin, line 83: This added line is also unnecessary

kcalvinalvin:
Tested. Looks good 👍

sloorush:
@adiabat @kcalvinalvin, this PR can be merged. 😅

kcalvinalvin:
> @adiabat @kcalvinalvin, this PR can be merged. 😅

I don't have merge access to this repo. @adiabat?

PR #329, opened 2021-10-25
accumulator/forest: Add AssertEqual() to Forest

kcalvinalvin:
Refactors the test function for checking if forests equal and makes it
public. Equal function is useful for testing and general sanity
checking.

### DIFF
--- a/accumulator/forest.go
+++ b/accumulator/forest.go
@@ -2,6 +2,7 @@ package accumulator
 
 import (
 	"encoding/binary"
+	"encoding/hex"
 	"fmt"
 	"os"
 	"sort"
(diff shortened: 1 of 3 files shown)


accumulator/forest.go
```diff
@@ -764,3 +765,89 @@ func (f *Forest) FindLeaf(leaf Hash) bool {
...
+	}
+
+	if f.numLeaves != compareForest.numLeaves {
+		err := fmt.Errorf("number of leaves aren't equal"+
+			"forest: %d, compareforest : %d\n", f.numLeaves,
+			compareForest.numLeaves)
+		return err
+	}
+
+	// If the sizes don't equal, return err.
```
kcalvinalvin, line 789: remove this.

devrandom (commented):
It seems to me that `Forest` could be enhanced to add a couple of functions that iterate on the rows and iterate on the elements of one row.  This would be while taking into account the number of rows and number of leaves - those determine the logical tree shape and extent.

accumulator/forest.go
```diff
@@ -764,3 +765,89 @@ func (f *Forest) FindLeaf(leaf Hash) bool {
...
+	// the f.size() only returns the count for the bottom row.
+	//
+	// The idea is that we'll keep moving up, and compare all the nodes.
+	var pos uint8
+	for h := uint8(0); h <= compareForest.rows; h++ {
+		// Calculate element count in the current row.
+		rowlen := uint8(1 << (compareForest.rows - h))
+
+		// Loop through all the elements in the current row.
+		for i := uint8(0); i < rowlen; i++ {
```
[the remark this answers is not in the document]
kcalvinalvin: Yup. Am changing it so you grab the logical rows with
```
logicalRows := treeRows(f.numLeaves)
```

There's other code that needs to be changed as well... I guess I should convert this to a wip

accumulator/forest.go
```diff
@@ -764,3 +765,89 @@ func (f *Forest) FindLeaf(leaf Hash) bool {
...
+				val, compVal)
+			return err
+		}
+	}
+
+	// Iterate through all the rows in the forest. We need to do this as
+	// the f.size() only returns the count for the bottom row.
+	//
+	// The idea is that we'll keep moving up, and compare all the nodes.
+	var pos uint8
```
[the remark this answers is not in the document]
kcalvinalvin: Good catch :)

accumulator/forest.go
```diff
@@ -764,3 +765,97 @@ func (f *Forest) FindLeaf(leaf Hash) bool {
...
+		// 12              13*
+		// |-------\       |-------\
+		// 08      09      10      11*
+		// |---\   |---\   |---\   |---\
+		// 00  01  02  03  04  05  06* 07*
+		fPos = getRowOffset(h, f.rows)
+		compPos = getRowOffset(h, compareForest.rows)
+
+		// Calculate element count in the current row.
+		elementCountAtRow := uint8(1 << (logicalRows - h))
```
devrandom, line 838: this is the physical length of the row, but there may be less items in it depending on `numLeaves`.  in the example you have above, the first row has 6 items instead of 8, the second has 3 instead of 4, etc.

accumulator/utils.go
```diff
@@ -338,11 +346,45 @@ func treeRows(n uint64) uint8 {
...
+//
+// Example: The below tree will return 1 as the logical number of rows is 1 for this
+// forest.
+//
+// row 2:
+//        |-------\
+// row 1: 04
+//        |---\   |---\
+// row 0: 00  01  02
+func logicalTreeRows(n uint64) uint8 {
```
adiabat, line 359: What's the difference between `treeRows()` and `logicalTreeRows()`?  `treeRows()` is a bit fancier but they do the same thing right?

kcalvinalvin: `treeRows()` will return the amount of rows that the forest will allocate for those number of leaves. `logicalTreeRows()` will return only the amount of rows that need to be there.

I put an example here with `numLeaves := uint64(3)`. `treeRows()` returns 2 because the forest will allocate enough for 2^2 leaves. But `logicalTreeRows()` returns 1 because that's where the highest root is.
https://play.golang.org/p/m-EWcl0aw0V

accumulator/forest.go
```diff
@@ -764,3 +765,97 @@ func (f *Forest) FindLeaf(leaf Hash) bool {
...
+		if !ok {
+			err := fmt.Errorf("miniHash %s doesn't exist in the the compared forest",
+				hex.EncodeToString(key[:]))
+			return err
+		}
+
+		if val != compVal {
+			err := fmt.Errorf("miniHash %s returned position %d for "+
+				"forest but %d for the compared forest", hex.EncodeToString(key[:]),
+				val, compVal)
```
adiabat, line 802: fwrewf

accumulator/forest.go
kcalvinalvin: Thanks. Will be changing to `f.numLeaves >> h`

accumulator/utils.go
kcalvinalvin: How about just using `uint8(bits.Len64(n) - 1)` in the code instead of making a new function? It's a single liner too.

accumulator/utils.go
```diff
@@ -338,11 +346,45 @@ func treeRows(n uint64) uint8 {
...
+// |---------------\
+// 12              13
+// |-------\       |-------\
+// 08      09      10      11
+// |---\   |---\   |---\   |---\
+// 00  01  02  03  04  05  06  07
+//
+// 8 = getRowOffset(1, 3)
+// 12 = getRowOffset(2, 3)
+func getRowOffset(row, forestRows uint8) uint64 {
```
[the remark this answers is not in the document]
kcalvinalvin: I can use `parentMany(0, row, forestRows)` for this PR. Should I leave `getRowOffset()` in `forestdata.go`?

accumulator/utils.go
adiabat: OK sure, that works.

PR #330, opened 2021-11-04
Fuzz forest / undo

devrandom:
TODO:

- [x] ensure that fuzzer job fails if crashes are found
- [x] DRY with existing SimChain

A simple fuzzer using [go-fuzz](https://github.com/dvyukov/go-fuzz).

- implement a fuzzer for forest / undo
- add a check that undo produces the same roots that were there before the block

This is probably not optimal - i.e. the execution time of each try could be reduced.

It also doesn't cover all of the branches after about 12 hours of CPU time - can be seen in the coverage.

Usage:

```shell
go install github.com/dvyukov/go-fuzz/go-fuzz@latest
go get github.com/dvyukov/go-fuzz/go-fuzz-dep
cd accumulator
go-fuzz-build
go-fuzz
```

Coverage:
```shell
go-fuzz -dumpcover
# `go tool cover` may fail due to a bug - you might need to remove a line that looks like:
# ...accumulator/forest.go:0.0,1.1 0 0
# as well as all lines that refer to files outside the repo
go tool cover -html=coverprofile
```

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -120,3 +120,69 @@ jobs:
       run: |
         cd $GOPATH/src/github.com/${{ github.repository }}
         ./test/check_proofs_backwards.sh utreexoserver "${{ github.workspace }}"
+
+  fuzz:
+    name: Fuzz
+    runs-on: ubuntu-latest
+    needs: Build
+    timeout-minutes: 5
+    steps:
+
+    - name: setup directory
+      shell: bash
+      run: |
+        mkdir -p "${{ github.workspace }}/fuzz-coverage"
+
+    # set GOPATH
+    - name: setup env
+      shell: bash
+      run: |
+        echo "${{ github.workspace }}/go/bin:" >> $GITHUB_PATH
+        echo "GOPATH=${{ github.workspace }}/go" >> $GITHUB_ENV
+
+    - name: Install Go
+      if: success()
+      uses: actions/setup-go@v1
+      with:
+        # go install doesn't work with 1.13
+        go-version: 1.17
+      id: go
+
+    - name: checkout
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 1
+        path: go/src/github.com/${{ github.repository }}
+
+    - name: Get dependencies
+      run: |
+        cd $GOPATH/src/github.com/${{ github.repository }}
+        go get -v -d ./...
+
+    - name: Install fuzzer
+      shell: bash
+      run: |
+        go install github.com/dvyukov/go-fuzz/go-fuzz@latest github.com/dvyukov/go-fuzz/go-fuzz-build@latest
+        go get github.com/dvyukov/go-fuzz/go-fuzz-dep
+
+    - name: Build
+      shell: bash
+      run: |
+        cd $GOPATH/src/github.com/${{ github.repository }}/accumulator
+        go-fuzz-build
+
+    - name: Fuzz
+      shell: bash
+      run: |
+        cd $GOPATH/src/github.com/${{ github.repository }}/accumulator
+        timeout --preserve-status --signal INT 120s go-fuzz -dumpcover |& tee fuzz.log
+        grep "crashers: 0," fuzz.log > /dev/null
+        sed -i '/0.0,1.1/d' coverprofile
+        cp coverprofile "${{ github.workspace }}/fuzz-coverage/"
+        go tool cover -html=coverprofile -o "${{ github.workspace }}/fuzz-coverage/cover.html"
+
+    - name: Fuzz coverage
+      uses: actions/upload-artifact@v2
+      with:
+        name: fuzz-coverage
+        path: "${{ github.workspace }}/fuzz-coverage"
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,9 @@ cmd/simcmd/simcmd
 cmd/utreexoclient/utreexoclient
 cmd/utreexoserver/utreexoserver
 cmd/utreexoclient/pollardFile 
+/accumulator/corpus/
+/accumulator/crashers/
+/accumulator/suppressions/
+/accumulator/accumulator-fuzz.zip
+/accumulator/coverprofile
+/accumulator/sonarprofile
--- a/accumulator/fuzz.go
+++ b/accumulator/fuzz.go
@@ -1,0 +1,73 @@
+//go:build gofuzz
+// +build gofuzz
+
+package accumulator
+
+import (
+	"bytes"
+	"fmt"
+	"reflect"
+)
+
+func undoOnceFuzzy(data *bytes.Buffer) error {
+	f := NewForest(RamForest, nil, "", 0)
+
+	seed0, err := data.ReadByte();
+	if err != nil { return nil }
+	seed1, err := data.ReadByte();
+	if err != nil { return nil }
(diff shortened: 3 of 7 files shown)


dergoegge:
Adding fuzzing is a great idea in my opinion! `go-fuzz` seems like a good pick for this, easy to use and it's easy to add new fuzzing tests. The bug with the coverfile is a bit annoying but there is a simple work around mentioned here: https://github.com/dvyukov/go-fuzz/issues/170#issuecomment-462297614 (`sed -i '/0.0,1.1/d' coverprofile`).

A github action for this could also be cool, to have this run at least for a short amount of time for new commits.

I tested this out and i am a bit surprised that the undo logic is working, because i was under the impression that it is currently broken. but maybe it isn't? cc @kcalvinalvin

devrandom:
I'll add an action, and maybe even product an artifact with the coverage

kcalvinalvin:
Hey, as mentioned in irc, I'll find time to review the PR. Not ignoring it and have actually skimmed through it few times.

>I tested this out and i am a bit surprised that the undo logic is working, because i was under the impression that it is currently broken. but maybe it isn't? cc @kcalvinalvin

The undo test is wrong in master. Below is the test where undo fails.

This snippet requires the forest.Equal() method in #329. You can just copy that Equal() function and paste it in forest.go (or anywhere else inside the package) to run this test.
```
func undoOnceRandom(blocks int32) error {
        f := NewForest(RamForest, nil, "", 0)
        f2 := NewForest(RamForest, nil, "", 0)

        sc := newSimChain(0x07)
        sc.lookahead = 0
        for b := int32(0); b < blocks; b++ {
                adds, durations, delHashes := sc.NextBlock(rand.Uint32() & 0x03)

                bp, err := f.ProveBatch(delHashes)
                if err != nil {
                        return err
                }
                ub, err := f.Modify(adds, bp.Targets)
                if err != nil {
                        return err
                }

                // undo every 3rd block
                if b%3 == 2 {
                        fBeforeString := f.ToString()
                        err := f.Undo(*ub)
                        if err != nil {
                                return err
                        }
                        sc.BackOne(adds, durations, delHashes)

                        err = f.Equal(f2)
                        if err != nil {
                                fmt.Println("--------------------------------------------------------------")
                                fmt.Println("forest not equal to the one before")
                                fmt.Println("Before undo")
                                fmt.Println(fBeforeString)

                                fmt.Println("After undo")
                                fmt.Println(f.ToString())

                                fmt.Println("f2")
                                fmt.Println(f2.ToString())
                                fmt.Println("--------------------------------------------------------------")
                                return err
                        }
                } else {
                        f2bp, err := f2.ProveBatch(delHashes)
                        if err != nil {
                                return err
                        }
                        _, err = f2.Modify(adds, f2bp.Targets)
                        if err != nil {
                                return err
                        }
                }
        }

        return nil
}
```

Undo basically doesn't "unadd" at the moment.

devrandom:
> The undo test is wrong in master. Below is the test where undo fails.

hmmm.... I wonder how that is possible, given the code on this branch to compare the roots succeeds:

https://github.com/mit-dci/utreexo/pull/330/files#diff-4258a9957941fc80b1c3f9c3e23f09967612084139427440cb0f493cc5e940edR151

kcalvinalvin:
> hmmm.... I wonder how that is possible, given code on this branch to compare the roots succeeds

It's because the `getRoots()` function is based off of `forest.numLeaves` and `forest.rows` and those things correctly get updated during an undo.

https://github.com/mit-dci/utreexo/blob/37699d1fcc9f59ebab4105fc92314842ccd7f495/accumulator/forest.go#L679-L692

However, the underlying forest didn't change. So the `getRoots()` function is grabbing non-roots. This actually wouldn't happen with `Pollard` because you keep updating the roots and don't calculate it on the fly.

Try:
```
func undoOnceRandom(blocks int32) error {
        f := NewForest(RamForest, nil, "", 0)

        sc := newSimChain(0x07)
        sc.lookahead = 0
        for b := int32(0); b < blocks; b++ {

                adds, durations, delHashes := sc.NextBlock(rand.Uint32() & 0x03)

                fmt.Printf("\t\tblock %d del %d add %d - %s\n",
                        sc.blockHeight, len(delHashes), len(adds), f.Stats())

                bp, err := f.ProveBatch(delHashes)
                if err != nil {
                        return err
                }
                beforeString := f.ToString()
                ub, err := f.Modify(adds, bp.Targets)
                if err != nil {
                        return err
                }

                // undo every 3rd block
                if b%3 == 2 {
                        fmt.Print(ub.ToString())
                        err := f.Undo(*ub)
                        if err != nil {
                                return err
                        }
                        afterString := f.ToString()

                        if beforeString != afterString {
                                err := fmt.Errorf("expected:\n %s\n but got:\n%s\n",
                                        beforeString, afterString)
                                panic(err)
                        }
                        sc.BackOne(adds, durations, delHashes)
                }

        }
        return nil
}
```

kcalvinalvin:
In general, I think adding fuzzing to the repository is good. However, I don't
think the approach taken in this PR is in the right direction.

The random bytes generated from go-fuzz is being fed into `undoOnceFuzzy()` but
that function isn't doing much with it.

The actual usage that I see from the randomness is in two places. Once when
generating the number of adds and once during generating the durations of each
add.

```
numAdds, err := data.ReadByte()
if err != nil {
        break
}
numAdds &= 0x1f
```

```
durations[j] = int32(s.rnd.Uint32() & s.durationMask)
```

Both of these don't really gain much from randomness. Randomness for the number
of utxos generated and the duration of each utxo doesn't really effect consensus
critical parts of the code.

Also, the the Undo() function isnt't a really great candidates for fuzzing since
Undo() only reads the undoblocks that was previously saved to the user's disk.
This can easily be tested by unit tests.

In general, functions for the pollard proof verification functions are better candidates for fuzz testing in this repo.
https://github.com/mit-dci/utreexo/blob/37699d1fcc9f59ebab4105fc92314842ccd7f495/accumulator/pollardproof.go#L7-L82

devrandom:
Here are my thoughts on the fuzzing approach.

This PR fuzzes the following:

- Forest manipulation - this is the basic data structure for bridge nodes, which is critical in the medium term as a critical component of the deployment.
- Undo - I think it's important to thoroughly test undo because reorgs don't happen that often, so any bugs might not be uncovered for a while otherwise.  I also don't think the logic is trivial enough to just depend on unit tests, especially since there may be interactions between the Forest transformation routines and Undo.

Note that although the name of the top-level routine is `undoOnceFuzzy` - the intention isn't that we are primarily testing Undo - it was just useful to start from the `undoOnceRandom` test because it exercises Forest too - and I kept the name.  It could be named `fuzzForestAndUndo`

The fuzzing entropy is used as follows:

- the number of adds in a block
- the duration of each UTXO, and therefore the number of dels in subsequent blocks

This generates forests with different shapes, and therefore exercises different aspects of the forest transformation logic.

I agree that Pollard should be tested too, but I'm leaving that for a follow-on effort.

kcalvinalvin:
Pondered a bit more about this and I think having this in the repo would be better than not having it in the repo. Is it ready to be merged?

devrandom:
@kcalvinalvin I will rebase and DRY the PR today

kcalvinalvin:
LGTM (sorry for the late review).

adiabat:
LGTM -- sorry for the huge delay; I was locked out of github.  Thanks!

PR #335, opened 2022-01-10
accumulator/forest: Export getroots function

kcalvinalvin:
Function getroots is exported so that bridge nodes are able to query
the accumulator roots.


adiabat:
Complying with 2022 capital requirements.

PR #338, opened 2022-02-10
accumulator/pollard: Pollard multiblockproof fix & utility functions

kcalvinalvin:
Pollard addition algorithm is modified to support multi-block proofs. New utility functions are also added for pollard.


adiabat:
the 
```
		n.remember = remember
```
Has more effects than it might seem: lots of things will be remembered and it might saturate the pollard.  
For now though this is the best we can do.  So basically, cool, merge, but
// TODO: fix remembers

Issue #340, opened 2022-02-18
Separate out the unit tests from the integration tests


kcalvinalvin:
Right now the repository states that all tests are failing which is not true. All the unit tests in the accumulator library are passing, the integration tests in `test/` are failing.

The github actions file could be modified so that the unit tests have a separate checkmark from the integration tests. The file at: https://github.com/mit-dci/utreexo/blob/master/.github/workflows/go.yml should be modified to achieve this.

samay-kothari:
Hi @kcalvinalvin, I am new to this project and want to contribute, this seems like a good first issue to me.
Can you please elaborate on how you need to distribute the tests.
What I can think right now is, we extract the unit test and integration test steps from the build job and run them as different jobs. Will this work?

samay-kothari:
Also fixed the test for test for the fuzz.

Refer: https://github.com/samay-kothari/utreexo/actions/runs/1904007542

kcalvinalvin:
Hey Samay, that looks good but the backwards integration test is failing because it's not finding any github artefacts.

```
Error: Unable to find any artifacts for the associated workflow
```

`Upload Testdata` should be done in the same step as the fetch iirc for it to work correctly.

I think that's the only thing that's missing. Could you make a PR for this and we can review it there.

samay-kothari:
Hi Calvin,
![Screenshot 2022-03-04 at 11 38 36 PM](https://user-images.githubusercontent.com/55805094/156818236-0168f7b7-8cee-4b60-a4db-43460642efba.png)
As you can see in this attached screenshot, the action of uplaod artifact is saying that there are no files at the provided path.
Since, the Unit tests and Integration Tests are seperated and Fuzz testing is fixed, I will make a pull request.

PR #341, opened 2022-03-04
Fixing the github workflow.

samay-kothari:
In reference to #340. Separated the unit tests and the integration tests. Fixed the fuzz testing.

Currently, the Build, Unit Tests, and Fuzz tests are working.

Tried to fix the backward integration tests, but there is some issue with the upload artifact action as it can't detect files at the provided path.


kcalvinalvin:
> Tried to fix the backward integration tests, but there is some issue with the upload artifact action as it can't detect files at the provided path.

Ah it may be because of the integration tests failing. @adiabat could you take a look

PR #352, opened 2022-04-15
accumulator/forestproofs: fix off-by-one error in `Prove` sanity check

theStack:
If the number of leaves is `numLeaves`, valid leaf positions are in the range of `(0, numLeaves-1)`, i.e. `pos == numLeaves` is already out of range. Take this into account for the sanity check before calling `detectSubTreeRows`.

### DIFF
--- a/accumulator/forestproofs.go
+++ b/accumulator/forestproofs.go
@@ -25,7 +25,7 @@ func (f *Forest) Prove(wanted Hash) (Proof, error) {
 	}
 
 	// should never happen
-	if pos > f.numLeaves {
+	if pos >= f.numLeaves {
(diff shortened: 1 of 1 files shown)


kcalvinalvin:
ACK 80599dc1bc04fe259dcd16266abdab5decad4fc6

adiabat:
Cool - thanks!  We should probably get rid of the Prove function since really we're always going to use batch proofs, even if there's only one leaf.

theStack:
> Cool - thanks! We should probably get rid of the Prove function since really we're always going to use batch proofs, even if there's only one leaf.

That makes sense. From a code reader's perspective it's very helpful though to see how generating a proof for only one leaf works first (at least for newbies like me) -- but of course that alone is hardly a good argument for keeping unused code around.

PR #354, opened 2022-04-18
accumulator/utils: remove unused function `subTreePositions`

theStack:
This function hasn't been used since commit
e83f377a2a6f9c3287ef64d01eab1e02b3c27606. It seems that it was only useful for removev2 and earlier, i.e. it makes sense to get rid of it.

### DIFF
--- a/accumulator/utils.go
+++ b/accumulator/utils.go
@@ -394,40 +394,6 @@ func getRootsForwards(leaves uint64, forestRows uint8, roots *[]uint64) []uint8 
 	return rows
 }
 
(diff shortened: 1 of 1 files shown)


kcalvinalvin:
ACK 1a5f7d17d41a096fd9ac1e290e545b4c86303d43

We're moving towards a non-swapping accumulator algorithm too it makes even more sense to get rid of this.

adiabat:
Looks good.  Yeah in the swapless branch lots of these types of functions are getting removed.

PR #355, opened 2022-04-19
accumulator/pollardutil: fix Pollard serialization size calculation

theStack:
The needed capacity for the Pollard serialization byte slice was underestimated (1 byte for each root instead of 32 bytes).
This is not the end of the world, but serialization most likely took longer than necessary due to necessary reallocation.


kcalvinalvin:
ACK 632bb3e35ad92235dfc39f7697af6575e120080a

kcalvinalvin:
ACK 632bb3e35ad92235dfc39f7697af6575e120080a

PR #356, opened 2022-04-19
accumulator/forestproofs: remove single-leaf `Prove{Many}`/`Verify{Many}` methods

theStack:
As suggested by @adiabat in https://github.com/mit-dci/utreexo/pull/352#issuecomment-1101462359:
> We should probably get rid of the Prove function since really we're always going to use batch proofs, even if there's only one leaf.

`Prove` and `Verify` are only used in one unit test, so this one is adapted to use the batch proof equivalents `ProveBatch` and `VerifyBatchProof` instead. Note that this also makes the structure `Proof` obsolete, so this is removed as well.

### DIFF
--- a/accumulator/forest_test.go
+++ b/accumulator/forest_test.go
@@ -195,7 +195,7 @@ func Test2Fwd1Back(t *testing.T) {
 		fmt.Printf(s)
 
 		// get proof for the first
-		_, err = f.Prove(adds[0].Hash)
+		_, err = f.ProveBatch([]Hash{adds[0].Hash})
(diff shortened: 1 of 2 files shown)


kcalvinalvin:
ACK 41978b67100a63881e0d68198a68a1bc0d9a849a

side-note: we should get rid of those commented out code like the one at https://github.com/mit-dci/utreexo/blob/6ac58e8b644714487118155f01240a4da4335fb8/accumulator/forest_test.go#L203-L210

Doesn't really serve much and hurts readability.

